PR #4, opened 2019-02-07
Add meter ratelimiting parameter in acl add function

softwarejl:
Add meter ratelimiting parameter in acl add function

### DIFF
--- a/examples/main.go
+++ b/examples/main.go
@@ -24,12 +24,11 @@ import (
 )
 
 const (
-	OVS_RUNDIR = "/var/run/openvswitch"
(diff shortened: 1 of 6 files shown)


examples/main.go
```diff
@@ -24,12 +24,11 @@ import (
 )
 
 const (
-	OVS_RUNDIR = "/var/run/openvswitch"
+	OVS_RUNDIR   = "/var/run/openvswitch"
```
hzhou8: Please split the format changes into a separate commit in this pull request

PR #9, opened 2019-03-07
Support Ovn Load Balancer operations

nikhildl12:
Support Ovn Load Balancer operations in go-ovn

### DIFF
--- a/examples/main.go
+++ b/examples/main.go
@@ -18,7 +18,6 @@ package main
 
 import (
 	"fmt"
-
 	"os"
 
 	"github.com/ebay/go-ovn/goovn"
@@ -34,6 +33,7 @@ const (
 
 var ovndbapi goovn.OVNDBApi
 
+
 func init() {
 	var ovs_rundir = os.Getenv("OVS_RUNDIR")
 	if ovs_rundir == "" {
@@ -44,6 +44,7 @@ func init() {
 }
 
 func main() {
+
 	ocmd := ovndbapi.LSWAdd("ls1")
 	ovndbapi.Execute(ocmd)
 	ocmd = ovndbapi.LSPAdd("ls1", "test")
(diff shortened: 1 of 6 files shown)


examples/main.go
```diff
@@ -97,4 +97,38 @@ func main() {
...
+	}
+	fmt.Printf("\n Lb found:%+v", lb[0])
+
+	fmt.Printf("\nDeleting LB")
+	ocmd = ovndbapi.LBDel("lb1")
+	err = ovndbapi.Execute(ocmd)
+	if err != nil {
+		fmt.Printf("err executing command:%v", err)
+		return
+	}
```
hzhou8: read again to make sure it is gone after delete.

goovn/apis.go
```diff
@@ -57,6 +57,13 @@ type OVNDBApi interface {
 	ASAdd(name string, addrs []string, external_ids map[string]string) *OvnCommand
 	// Delete addressset
 	ASDel(name string) *OvnCommand
+	// Add LB
+	LBAdd(name string, vipPort string, protocol string, addrs []string) *OvnCommand
+	// Delete LB with given name
+	LBDel(name string) *OvnCommand
+	// Update existing LB
+	LBUpdate(name string, vipPort string, protocol string, addrs []string) *OvnCommand
```
hzhou8: Is it better to add LBUpdateMembers() to support more efficient updates when the size of members are big? This interface can support passing added and deleted members only, without re-providing unchanged members. (Of course, it doesn't have to be in this patch. It can be added later)

examples/main.go
hzhou8: BTW, it is better to add tests in ovnnbimp_test.go, so that we can run "go test".

softwarejl:
not sure if need event feedback logical for LB vip creation?

vtolstov:
does it ready for merge? i want to send own pr for fatal and return error codes...

hzhou8:
@vtolstov I requested a test case update. It should be a quick update and will be merged soon. However, you send your PR whenever it is ready so that people can review as early as they can (and probably you need to rebase later if there is a conflict).

nikhildl12:
@hzhou8: Moved the tests to ovnnbimp_test.go

goovn/apis.go
nikhildl12: Good suggestion. I will handle it in a follow up patch

PR #14, opened 2019-03-11
return errors when it happened

vtolstov:

### DIFF
--- a/goovn/ovndb.go
+++ b/goovn/ovndb.go
@@ -17,9 +17,7 @@
 package goovn
 
 import (
-	"errors"
 	"fmt"
-	"os"
 	"sync"
 
 	"github.com/socketplane/libovsdb"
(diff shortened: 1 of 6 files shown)


softwarejl:
Can we use the V2 interface to compatible with the previous one?

hzhou8:
Thanks @vtolstov ! I haven't looked every details yet but overall it looks good to me. I agree with @softwarejl on the compatibility concern. There are two ways to handle it. 1) As mentioned by @softwarejl , use V2 interface. 2) We may branch and release version 0.1 before merging this. Both approaches have pros and cons. For option 1), we will need to maintain 2 versions of interfaces for certain period. For option 2), old interface users won't get new features/bug fixes unless those are backported to v0.1 branch. Any thoughts on this?

vtolstov:
what do you think about go.mod ?

vtolstov:
I'm preferred option 2. So if you need the new features expected to have some breaks....

vtolstov:
after that pr, I'm plan to push dhcp support, and after some fixes logical router support

hzhou8:
@vtolstov thanks for the updates. I'll let @softwarejl do a detailed review. BTW, I saw some changes unrelated to what the patch was described to do "return errors when it happened", such as the changes for GetInstance(). If it is needed, it is better to be splitted as a separate commit. It is preferred that one commit fixes one problem.

For the compatibility problem, I think we agreed with option 2) branch and release version 0.1, and put this change as v0.2. I will do the branching.

goovn/ovndb.go
```diff
@@ -75,31 +73,28 @@ type OVNDB struct {
 	imp *ovnDBImp
 }
 
-var once sync.Once
-var ovnDBApi OVNDBApi
```
softwarejl: why we remove that?

softwarejl: we must keep only one instance in one process.

goovn/uuid.go
```diff
@@ -0,0 +1,13 @@
+package goovn
+
+import (
+	"github.com/google/uuid"
+)
+
+func newUUID() (string, error) {
```
softwarejl: what is this for?

goovn/ovndb.go
vtolstov: Ah, ok.. I'm prefer to use client like semantics..
Does it strict requirement to have single instance?

goovn/uuid.go
vtolstov: I'm change from rand to proper uuid, i think that is more generic and have less ability to duplicate..

vtolstov:
Ok im revert methid name GetInstance

vtolstov:
im return ol method name and once

vtolstov:
Does i need to change someting else? (i'm plan to use this branch to logical router and dhcp)

vtolstov:
Thanks

PR #17, opened 2019-03-12
Adding SSL support for OVN

noah8713:


softwarejl:
do we really need to upgrade vendor?

noah8713:
I think it would be good one time thing since there are couple of changes in latest vendor codes. I can submit that as a seperate patch. @hzhou8 are you fine with vendor edit or want to upgrade vendor?

hzhou8:
@noah8713 the vendor socketplane/libovsdb hasn't been updated for 2 years. We were update to date since day 1 and even have bug fixes on top of that. The files you added were deleted on purpose by @sudhi-vm 's commit (https://github.com/eBay/go-ovn/commit/340f8e3d3c26e15381cdd369ee9a0c02b374be02). If you think that is a mistake please confirm with @sudhi-vm and fix in separate commit but don't mix with this one.

noah8713:
Sure @hzhou8 I will talk to @sudhi-vm and confirm the intent of that PR. For now I did inline edit in the same vendor folder because there was not vendor updates on this repo since long for ssl to work with go-ovn as per  PR https://github.com/socketplane/libovsdb/pull/69 . Hence, either we do edit of vendor to support this patch or update vendore after https://github.com/socketplane/libovsdb/pull/69 is merged so that libovsdb is upto date. Fine with any options.

hzhou8:
Yes, you can make changes in vendor, but one commit for one change. Please don't mix with anything unrelated.

noah8713:
@hzhou8 as discussed,   as maintenance of upstream repo socketplane is unknown, I have edited the vendor files directly for ssl changes. PTAL.

vtolstov:
May be fork it and use directly from ebay?

ср, 13 мар. 2019 г., 20:27 aginwala <notifications@github.com>:

vendor/github.com/socketplane/libovsdb/client.go
```diff
@@ -82,7 +142,7 @@ func ConnectUsingProtocol(protocol string, target string) (*OvsdbClient, error)
 }
 
 // Connect creates an OVSDB connection and returns and OvsdbClient
-func Connect(ipAddr string, port int) (*OvsdbClient, error) {
+func Connect(ipAddr string, port int, protocol string) (*OvsdbClient, error) {
```
nikhildl12: Add comments for this function regarding supported protocols and which envs are being used for cert files, so that the caller of this library knows about it

noah8713: done

softwarejl:
LGTM

PR #21, opened 2019-03-14
add dhcp options support

vtolstov:


vtolstov:
@hzhou8 sorry for question, if i need to implement 
```
dhcp-options-set-options dhcp-option [key=value]...
              Set the DHCP Options for the dhcp-option UUID.
```

i'm already have created row in DHCP_Options table, as i understand func to update options must receives uuid of option, does i need mutate or update option to this row?

vtolstov:
or @softwarejl do you know answer for my question?

vtolstov:
Near completed

vtolstov:
@hzhou8 what you think about analog of lsp-get-dhcpv4-options ? does it need to return only uuid of dhcp option or full option?
I think about to return full struct (with uuid and map of options and cidr)

vtolstov:
@hzhou8 can you say, how this package library knows about new rows? I see populateCache, but i don't see that any new items added to map when update happened.

vtolstov:
ah i see, odbi.cache[table][uuid] = row.New

vtolstov:
@hzhou8 @softwarejl can you look at this?

vtolstov:
needs #28 before merging

vtolstov:
@hzhou8 cay you merge this?via squash commits?

PR #22, opened 2019-03-14
use go.mod and move to topdir

vtolstov:
* move all goovn files to topdir
* fix broken test (missing -e in script, so tests always passed)
* bump go in travis
* use go modules and vendor all deps


vtolstov:
@hzhou8 @softwarejl i see that test is broken because it not returns error status, i'm fix test in this commit and now i don't understand how cache is working? i don't see any function that fills cache when i create logical switch or something else. So methods that works with this caches fails

vtolstov:
cache problem solved adding c.SetBlocking(true) in libovsdb/client.go back (broken after ssl merging)

vtolstov:
does somebody can helps me fix present test issues? this is acl tests and deletion of load balancer (i don't understand why load balancer deletion not delete it)

vtolstov:
@hzhou8 @softwarejl pr is ready to review, sorry it have many changes, but i think that now code looks really better, and also fixed tests can detect breaking changes.

hzhou8:
> cache problem solved by adding c.SetBlocking(true) in libovsdb/client.go back (broken after ssl merging)
@noah8713 @nikhildl12 

Thanks @vtolstov Let's fix the broken things in separate patches if they are not introduced by current PR. BTW, why wasn't it detected by CI when the original PR was submitted? Was it missing tests?

vtolstov:
test script does not have #!/bin/sh -e so it not stop on error and last command in script always successed

vtolstov:
I can put separate patch for fix test script and may be separate patch for blocking mode to rpc2. What you say?

vtolstov:
@hzhou8 #27

hzhou8:
> test script does not have #!/bin/sh -e so it not stop on error and last command in script always successed
Oh, my bad, horrible mistake. I did test the failure cases but didn't test it after adding the cleanup commands after the "go test". Thanks for fixing.

(Please, it is always better to separate patches for different problem fixes.)

vtolstov:
i'm prepare patches, thanks

vtolstov:
i'm fix this pr after other pr are merged (because this is trivial and mostly go mod stuff)

PR #27, opened 2019-03-15
fix tests and add go 1.12.x to travis testing

vtolstov:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,9 @@
 language: go
 
+go:
+  - 1.11.x
+  - 1.12.x
+
 os:
   - linux
 
--- a/.travis/test.sh
+++ b/.travis/test.sh
@@ -1,7 +1,9 @@
+#!/bin/sh -ex
+
 rm -rf ovs
 git clone https://github.com/openvswitch/ovs.git
 cd ovs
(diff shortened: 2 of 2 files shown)


vtolstov:
test is expected to fail, as you see my commit does not have vendor dep, so it fails.
next pr fix monitor and vendor dep

.travis/test.sh
```diff
@@ -104,6 +106,7 @@ The backup database file is sandbox/${db}2.db
 
 ovn_start_db nb standalone 1 $srcdir/ovn/ovn-nb.ovsschema
 
-cd ../goovn; go test
+export GO111MODULE=on
```
hzhou8: Please keep the go mod related changes in the go mod patch, instead of this one.

.travis.yml
```diff
@@ -1,5 +1,8 @@
 language: go
 
+go:
+  - 1.12.x
```
hzhou8: why bump the version? Is it required because of go mod? If so, please move this change to the "go mod" PR, instead of this one. This PR is supposed to fix the test.sh bug.

vtolstov:
thanks for review, i'm add to travis go 1.12.x and go 1.11.x, and fix go.mod related stuff

PR #28, opened 2019-03-15
build and test improvements

vtolstov:


vtolstov:
after merging pr #27 this can be merged to see travis ci results

hzhou8:
This PR is titled "fix vendor". Could you put more details what is broken?

I see you updated all vendor packages to use latest version, but I don't see the reason for updating vendor. In general I'd not use latest versions for vendor packages, because of reliability concerns. We can update vendor versions when we see a critical bug fix or major improvements. Or we can update them to a new version after testing with confidence, but it is better to always pin to a tag/commit for better version tracking. What do you think?

vtolstov:
i'm fix all tests and code. now looks good.
also i set clone depth = 1 when cloning ovs, that speedup build.

vtolstov:
@hzhou8 @softwarejl can you look at this, when have time?

hzhou8:
It is great to see the CI fixed. Thanks!

However, this commit contains many changes that are not directly related. I saw at least 3 different changes:

1. Fix a bug in aclAddImp()
2. Fix test cases (that is not related to 1))
3. Update vendors
4. Improve test with git clone --depth=1

Could you please split them into 4 different PRs, each solve one problem, so that they can be reviewed and merged more easily, also showing up more clearly in commit history?

Regarding 3) the vendor update, I posted some comments but I still didn't see your feedback:

> I see you updated all vendor packages to use latest version, but I don't see the reason for updating vendor. In general I'd not use latest versions for vendor packages, because of reliability concerns. We can update vendor versions when we see a critical bug fix or major improvements. Or we can update them to a new version after testing with confidence, but it is better to always pin to a tag/commit for better version tracking. What do you think?

vtolstov:
i think that this all commits related to test, so i combine all of them.
about vendor updates, i plan to add go module support in next commit and freeze versions inside it.

vtolstov:
and file vendor/modules.txt not used now, because of build and test not use vendor module now

hzhou8:
> i think that this all commits related to test, so i combine all of them.

They are not all related to test, e.g. 1) and 3). Even if they were all related to test, they should still be committed separately if they are for different problems in test. "git reset" and "git add -i" is quit convenient for splitting commits.

2) and 4) look good enough for me and I want to merge them ASAP if you could simply split them out.

vtolstov:
i'm split to small commits, can you look now?

vtolstov:
for google/uuid i'm use latest released version commit

vtolstov:
also i'm not update all deps

goovn/uuid.go
```diff
@@ -9,5 +11,5 @@ func newUUID() (string, error) {
 	if err != nil {
 		return "", err
 	}
-	return id.String(), nil
+	return "row" + strings.Replace(id.String(), "-", "_", -1), nil
```
hzhou8: The function name is very generic, so it is not appropriate to return a string with prefix "row".

hzhou8: I don't understand how this commit is related to ovn-nbctl behavior. Could you explain a little more? I see the function newUUID() is used only to generate an named ID (which doesn't have to be an uuid format) when doing insert operations. The name is useful only during the transaction execution when a new row inserted in the transaction is referenced by another operation in the same transaction.

vtolstov: I see that if named uuid have string like uuid - record not inserted.
This also applied to case then the strings contains _ but, i'm check ovn-nbctl request and see that it sends in named uuid - string with row prefix and uuid with _, so i mimic this behavior.

vtolstov: for func name i can change it to newRowUUID for example?

vtolstov:
i'm fix commit messages and uuid.go

vtolstov:
Great,thanks!

PR #31, opened 2019-03-21
additional checks and minor speedup

vtolstov:


vtolstov:
i'm pull upstream and merge it

vtolstov:
@softwarejl what you say?

vtolstov:
and cleanup vendor dir, all deps are fetched when building package and vendor dir only needed in final product that uses this package

hzhou8:
@vtolstov thanks for the improvements! I appreciate that you have separate commits for different problems, but I still recommend in the future using different git branches for logically unrelated and independent changes, and sending separate PRs, which can speed up PR merging. Otherwise, one commit in the PR could block the whole PR's merging. (However, it is completely ok and recommended that multiple commits that are closely related and depended by each other go with single PR)

For this PR, the CI is failing, could you take a look?

vtolstov:
> @vtolstov thanks for the improvements! I appreciate that you have separate commits for different problems, but I still recommend in the future using different git branches for logically unrelated and independent changes, and sending separate PRs, which can speed up PR merging. Otherwise, one commit in the PR could block the whole PR's merging. (However, it is completely ok and recommended that multiple commits that are closely related and depended by each other go with single PR)
> 
> For this PR, the CI is failing, could you take a look?

Yes, this is trivial fix, i'm already do it in qos branch, but forget to get it here.

vtolstov:
@softwarejl thanks for suggestion, i'm try to follow it.

vtolstov:
@softwarejl all tests passed now

.travis/test.sh
```diff
@@ -105,7 +105,9 @@ The backup database file is sandbox/${db}2.db
 }
 
 ovn_start_db nb standalone 1 $srcdir/ovn/ovn-nb.ovsschema
-
-cd ../; go test -v
+export GO111MODULE=on
+cd ../;
```
hzhou8: nit: ';' not needed

vtolstov:
@softwarejl @hzhou8 i think this is fine now, can you look and merge?

apis.go
```diff
@@ -94,23 +94,25 @@ type OVNDBApi interface {
 	Execute(cmds ...*OvnCommand) error
 
 	// Get all logical switches
-	GetLogicSwitches() []*LogicalSwitch
+	GetLogicalSwitches() ([]*LogicalSwitch, error)
```
hzhou8: I agree that GetLogicalSwitches is better. Just to make sure @lebauce won't be surprised of this API change.

ovnnbimp.go
```diff
@@ -88,9 +94,15 @@ func (odbi *ovnDBImp) oMapContians(s, t map[interface{}]interface{}) bool {
 }
 
 func (odbi *ovnDBImp) getRowUUIDContainsUUID(table, field, uuid string) (string, error) {
-	odbi.cachemutex.Lock()
-	defer odbi.cachemutex.Unlock()
-	for id, drows := range odbi.cache[table] {
+	odbi.cachemutex.RLock()
+	defer odbi.cachemutex.RUnlock()
+
+	cacheTable, ok := odbi.cache[table]
```
hzhou8: If table doesn't exist, doesn't it mean there is something fundamentally wrong? E.g. schema problem? Returning ErrorNotFound seems not appropriate for this case.

vtolstov: O think that i can create special error for such case

vtolstov:
i'm rebase, and create new error that can be used to return, when table or item in table not found (useful to internal function like rowToXXX , because they used when mutex locked and uuid already getted via range iterator.

apis.go
vtolstov: i don't think that this is problem, because this is already new version, so we don't break old.

PR #32, opened 2019-03-21
Fix unit tests as error handling commit missed to address it

noah8713:


vtolstov:
i think that this can be fixed by https://github.com/eBay/go-ovn/pull/31

noah8713:
@vtolstov Thanks for pointer. Better to use separate commit for this fix. Each merge should not break the functionality. Lets merge this and then you can rebase on top of this.

vtolstov:
main problem that this example not tested by script, may be add go build -o /dev/null this file in tests?

vtolstov:
so if something breaks in api we can known?

noah8713:
Yeah sure in the travis ci script. We can also add to test examples.go. However, lets use a separate commit for that too

vtolstov:
i think that pr #31 that have commits for each stuff is fine (yes it lack test for examples)

vtolstov:
and also nobody restrict to revert one of commits later if that needed

PR #33, opened 2019-03-21
Fix LB delete operation for all ovn versions:

noah8713:
Creating LB will add load_balancer entries in all logical switches in ovn
to ensure routing is taken care within all the lswitches for the particular vip.
However, delete lb doesnt iterate to all the lswitches and hence lb
delete fails saying multiple references found accross multiple switches.
Hence, we need to clean up lbuuid from all lswitches to avoid this error.
Newer versions of OVN do handle this condition by default where delete LB
will delete all the references too. So, adding this support for environments
using older ovn versions

### DIFF
--- a/load_balancer.go
+++ b/load_balancer.go
@@ -109,13 +109,45 @@ func (odbi *ovnDBImp) lbAddImp(name string, vipPort string, protocol string, add
 }
 
 func (odbi *ovnDBImp) lbDelImp(name string) (*OvnCommand, error) {
+	var operations []libovsdb.Operation
+
 	condition := libovsdb.NewCondition("name", "==", name)
 	deleteOp := libovsdb.Operation{
 		Op:    opDelete,
 		Table: tableLoadBalancer,
 		Where: []interface{}{condition},
 	}
-	operations := []libovsdb.Operation{deleteOp}
+
+	// Also delete references from Logical switches
+	row := make(OVNRow)
+	row["name"] = name
+	lbuuid := odbi.getRowUUID(tableLoadBalancer, row)
+	if len(lbuuid) == 0 {
+		return nil, ErrorNotFound
+	}
+	mutateUUID := []libovsdb.UUID{{lbuuid}}
+	mutateSet, err := libovsdb.NewOvsSet(mutateUUID)
+	if err != nil {
+		return nil, err
+	}
+	mutation := libovsdb.NewMutation("load_balancer", opDelete, mutateSet)
+	lswitches, err := odbi.getRowsMatchingUUID(tableLogicalSwitch, "load_balancer", lbuuid)
+	if err != nil && err != ErrorNotFound {
+		return nil, err
+	} else if err == nil {
+		// mutate all matching lswitches for the corresponding load_balancer
+		for _, lswitch := range lswitches {
+			mucondition := libovsdb.NewCondition("_uuid", "==", libovsdb.UUID{lswitch})
+			mutateOp := libovsdb.Operation{
+				Op:        opMutate,
+				Table:     tableLogicalSwitch,
+				Mutations: []interface{}{mutation},
+				Where:     []interface{}{mucondition},
+			}
+			operations = append(operations, mutateOp)
+		}
+	}
+	operations = append(operations, deleteOp)
 	return &OvnCommand{operations, odbi, make([][]map[string]interface{}, len(operations))}, nil
 }
 
--- a/ovnnbimp.go
+++ b/ovnnbimp.go
@@ -133,6 +133,22 @@ func (odbi *ovnDBImp) getRowUUIDContainsUUID(table, field, uuid string) (string,
 	return "", ErrorNotFound
 }
 
+func (odbi *ovnDBImp) getRowsMatchingUUID(table, field, uuid string) ([]string, error) {
+	odbi.cachemutex.Lock()
+	defer odbi.cachemutex.Unlock()
+	var uuids []string
+	for id, drows := range odbi.cache[table] {
+		v := fmt.Sprintf("%s", drows.Fields[field])
+		if strings.Contains(v, uuid) {
+			uuids = append(uuids, id)
+		}
+	}
+	if len(uuids) == 0 {
+		return uuids, ErrorNotFound
+	}
+	return uuids, nil
+}
+
 func (odbi *ovnDBImp) transact(ops ...libovsdb.Operation) ([]libovsdb.OperationResult, error) {
 	// Only support one trans at same time now.
 	odbi.tranmutex.Lock()
--- a/ovnnbimp_test.go
+++ b/ovnnbimp_test.go
@@ -458,7 +458,7 @@ func TestLoadBalancer(t *testing.T) {
 
 	t.Logf("Deleting LB")
 	ocmd, err = ovndbapi.LBDel("lb1")
-	if err != nil {
+	if err != nil && err != ErrorNotFound {
 		t.Fatal(err)
 	}
 	err = ovndbapi.Execute(ocmd)


noah8713:
@nikhildl12 : PTAL

noah8713:
@vtolstov: @softwarejl 
 Are tests running against master branch always in CI? The above test fails on dhcp vs lb object not found `
LM-SJC-11015761:go-ovn aginwala$ go test -v
2019/03/21 16:40:45 client: connected to:  192.168.56.101:6641
=== RUN   TestDHCPOptions
--- FAIL: TestDHCPOptions (0.00s)
    dhcp_options_test.go:33: object exist
=== RUN   TestLogicalRouterPort
--- PASS: TestLogicalRouterPort (0.01s)
=== RUN   TestLogicalRouter
--- PASS: TestLogicalRouter (0.00s)
=== RUN   TestACLs
--- FAIL: TestACLs (0.00s)
    ovnnbimp_test.go:91: object exist
=== RUN   TestAddressSet
--- PASS: TestAddressSet (0.01s)
=== RUN   TestLoadBalancer
--- PASS: TestLoadBalancer (0.00s)
    ovnnbimp_test.go:388: Adding LB to OVN
    ovnnbimp_test.go:397: Adding LB to OVN Done
    ovnnbimp_test.go:399: Updating LB to OVN
    ovnnbimp_test.go:408: Updating LB to OVN done
    ovnnbimp_test.go:410: Gettting LB by name
    ovnnbimp_test.go:415: Lb found:&{UUID:52e9b312-8ad0-40b3-883f-b1e9f84015bd Name:lb1 vips:map[192.168.0.10:80:10.10.10.127:8080,10.10.10.120:8080] protocol:tcp ExternalID:map[]}
    ovnnbimp_test.go:417: Deleting LB
    ovnnbimp_test.go:432: LB deletion done
FAIL
exit status 1`

load_balancer.go
```diff
@@ -118,7 +118,34 @@ func (odbi *ovnDBImp) lbDelImp(name string) (*OvnCommand, error) {
...
+	mutateSet, err := libovsdb.NewOvsSet(mutateUUID)
+	if err != nil {
+		return nil, err
+	}
+	mutation := libovsdb.NewMutation("load_balancer", opDelete, mutateSet)
+	ucondition, err := odbi.getRowUUIDContainsUUID(tableLogicalSwitch, "load_balancer", lbuuid)
+	if err != nil {
+		return nil, err
+	}
+	mucondition := libovsdb.NewCondition("load_balancer", "==", libovsdb.UUID{ucondition})
```
softwarejl: we can delete like that?  in this case, should update loadbalance array in logical switch.

softwarejl: this will delete all LB in LP list ? are you sure?

softwarejl: could you write a test case for that, I can review.

vtolstov: @noah8713 local test i ran before create pr. Main problem that if test fails, cleanup phase not running, so if delete dhcp options for example not running, second iteration of test run always says that object exists.
As for load balancer may be we need you fix for all versions of ovn, because sometimes local test run fails for me with the same error, that load balancer not deleted.

vtolstov:
```
diff --git a/ovnnbimp_test.go b/ovnnbimp_test.go
index 8fb47af75010..08803c8957e1 100644
--- a/ovnnbimp_test.go
+++ b/ovnnbimp_test.go
@@ -468,7 +468,7 @@ func TestLoadBalancer(t *testing.T) {
 
        // Verify deletion
        lb, err = ovndbapi.GetLB("lb1")
- if err != nil {
+       if err != nil && err != ErrorNotFound {
                t.Fatal(err)
        }
        if len(lb) != 0 {
```

vtolstov:
the above fix needed to test case or it fails always

vtolstov:
also in my case tests fails for load balancer with this patch too (it fails before this patch, load balancer not deleted)

vtolstov:
@noah8713 i'm recheck load balancer creation, attaching it to ls and deletion. i  don't know how in old ovn versions, but in 2.10 creating lb not creates it in all ls, but only creates in load_balancer table, to add it to switch you need to add it to switch via additional command, after that you can't delete lb before you delete it from logical_switch. so i don't think that you pr is right.
We need to have separate method to attach/detach lb from ls.

hzhou8:
@vtolstov The background of this PR is: in latest OVN upstream, reference between LogicalSwitch and LoadBalancer changed from strong reference to weak reference. In older versions, since it is strong reference, LB can't be deleted if it is still used in any LS, which is inconvenient. In new version, since it is weak reference, deleting LB will automatically deleting the references from all LSes that are using it. The purpose of this PR is to provide similar convenience to clients even when running with older OVN.

@noah8713 As pointed by @softwarejl the implementation has problem (although the idea is correct). Please update and add test cases to ensure correctness. Thanks a lot.

vtolstov:
I think that i fix this pr (@noah8713 sorry, if you want to fix it yourself)
Logic is simple, if uuid in LogicalSwitch table of load balancer not found, not create mutation on this table)

vtolstov:
@softwarejl please, check

vtolstov:
@noah8713 can you check this pr too? I'm commit to you pr branch

noah8713:
@vtolstov : Sorry for the delay. Thanks for the suggestions/improvements in the PR. Changes seems reasonable and makes sense. I tested the same and have amended them along with changes in UT. Also changed the commit message to ensure compatibility for all ovn version. PTAL

noah8713:
> Request changes

Done. PTAL at revised commit!

noah8713:
> @noah8713 i'm recheck load balancer creation, attaching it to ls and deletion. i don't know how in old ovn versions, but in 2.10 creating lb not creates it in all ls, but only creates in load_balancer table, to add it to switch you need to add it to switch via additional command, after that you can't delete lb before you delete it from logical_switch. so i don't think that you pr is right.
> We need to have separate method to attach/detach lb from ls.

For attach, we might need to do the same as TODO in new PR. e.g. to use the LS as a LB with a flag or something ; only then mutate LS table,

load_balancer.go
noah8713: @softwarejl : It will only delete LB by matching uuid https://github.com/eBay/go-ovn/pull/33/files#diff-9a905803b6ff1b7d0e209a45f0ae354fR139 and not all

vtolstov:
> For attach, we might need to do the same as TODO in new PR. e.g. to use the LS as a LB with a flag or something ; only then mutate LS table,

Can you create issue ?

noah8713:
> > For attach, we might need to do the same as TODO in new PR. e.g. to use the LS as a LB with a flag or something ; only then mutate LS table,
> 
> Can you create issue ?
There are multiple approaches here: 
1. addLB should just create LB and nothing else
2. AddLB should take array of lswitches too so that create LB will also attach them to Lswitches(add + mutate). e.g. 
ovn-nbctl set Logical_Switch <ls_uuid> load_balancer=<lbuuid>. 
3. Keep AddLB as independent and add new method setLBLSwitch which just mutates LSwitch table.

Let me know which one we should take up and open the issue accordingly. 
cc @hzhou8

vtolstov:
i'm prefer to mimic ovn-nbctl commands now.

load_balancer.go
```diff
@@ -109,13 +109,42 @@ func (odbi *ovnDBImp) lbAddImp(name string, vipPort string, protocol string, add
...
+	mutateSet, err := libovsdb.NewOvsSet(mutateUUID)
+	if err != nil {
+		return nil, err
+	}
+	mutation := libovsdb.NewMutation("load_balancer", opDelete, mutateSet)
+	ucondition, err := odbi.getRowUUIDContainsUUID(tableLogicalSwitch, "load_balancer", lbuuid)
+	if err != nil && err != ErrorNotFound {
+		return nil, err
+	} else if err == nil {
+		mucondition := libovsdb.NewCondition("load_balancer", "==", libovsdb.UUID{ucondition})
```
hzhou8: ucondition is a row uuid in LogicalSwitch table. How come it is compared with "load_balancer" field? Please add test cases to make sure the code actually works as expected.

vtolstov: i think that this is error, because getRowUUIDContainsUUID returns uuid of row in logical switch table, so we need to use lbuuid in mutation and delete it from load_balancer row in logical switch table.

ovnnbimp.go
```diff
@@ -133,6 +133,22 @@ func (odbi *ovnDBImp) getRowUUIDContainsUUID(table, field, uuid string) (string,
 	return "", ErrorNotFound
 }
 
+func (odbi *ovnDBImp) getRowsMatchingUUID(table, field, uuid string) ([]string, error) {
+	odbi.cachemutex.Lock()
+	defer odbi.cachemutex.Unlock()
+	var uuids []string
+	for id, drows := range odbi.cache[table] {
+		v := fmt.Sprintf("%s", drows.Fields[field])
```
softwarejl: why can not add this condition in delete transaction directly?

load_balancer.go
```diff
@@ -109,13 +109,42 @@ func (odbi *ovnDBImp) lbAddImp(name string, vipPort string, protocol string, add
...
+	if len(lbuuid) == 0 {
+		return nil, ErrorNotFound
+	}
+	mutateUUID := []libovsdb.UUID{{lbuuid}}
+	mutateSet, err := libovsdb.NewOvsSet(mutateUUID)
+	if err != nil {
+		return nil, err
+	}
+	mutation := libovsdb.NewMutation("load_balancer", opDelete, mutateSet)
+	ucondition, err := odbi.getRowUUIDContainsUUID(tableLogicalSwitch, "load_balancer", lbuuid)
```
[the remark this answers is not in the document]
noah8713: Done.

load_balancer.go
noah8713: @hzhou8 corrected and verified. Sorry it was _uuid. PTAL at revised PR.

ovnnbimp.go
noah8713: for simplicity. Easier for review and understanding the logic

vtolstov:
LGTM

Issue #29, opened 2019-03-19
external_ids support when creating items


vtolstov:
What you position about api in case of external_ids ? As i see ovn-nbctl does not provide in command line ability to set external_ids, but in case of api case this is really needed feature, to allow link internal own db id/uuid with ovn item.

Now i'm think about modify api and all functions that modify/create/delete items receives not ovn-nbctl options, but pointer tu needed struct.
So instead of have lswAddImp(lsw string) do lswAddImp(lsw *LogicalSwitch)
marshal , after that it can be converted to map[string]interface{}

vtolstov:
we can use https://godoc.org/github.com/mitchellh/mapstructure to convert from map[string]interface

vtolstov:
@hzhou8 any suggestions?

vtolstov:
@softwarejl what do you think about interface changes? another thing not pass struct is use option like feature.
So for each ovn table we have item Option, and passing it to Delete, Create, Update methods, slice of such options set struct row items.
Drawback for this - for each uniq item in struct we need to have different option. So this is many options.

vtolstov:
@hzhou8 @softwarejl i need you approve and suggestions. This is not small task, but it don't need too much time to provide WIP for this.

hzhou8:
@vtolstov I support this idea! @softwarejl and I had discussed about having more generic API interface many times, just didn't get bandwidth to work on it. Maybe you could have a POC first. @softwarejl what do you think?

softwarejl:
Yes, we need to improve on that part, beside the interface, we need a way to define the data structure which aligns  with ovn table structure. we need a poc firstly.

vtolstov:
first change in this series #42

vtolstov:
@hzhou8 this is first part of storing in cache needed struct https://github.com/eBay/go-ovn/pull/44/files#diff-48d527baa1a7dc53e558c5548785cf1bR277

vtolstov:
i think that if we store needed type inside cache all thing can be very simple.
next sep is write helper to check if some struct holds in another (so we can pass LogicalSwitch with only uuid setted and get back full row from cache if it exists)

vtolstov:
if somebody want helps me , i'l be happy =)

hzhou8:
[...]

PR #34, opened 2019-03-22
qos support

vtolstov:


qos.go
```diff
@@ -114,25 +114,47 @@ func (odbi *ovnDBImp) addQoSImp(ls string, direction string, priority *int, matc
 }
 
 func (odbi *ovnDBImp) delQoSImp(ls string, direction string, priority *int, match string) (*OvnCommand, error) {
```
softwarejl: why using point?

vtolstov:
@softwarejl pointer used to distinguish 0 vs unset param. Without it we always need to create rule with priority 0 if user not specify, i don't know what right solution....

vtolstov:
> @softwarejl pointer used to distinguish 0 vs unset param. Without it we always need to create rule with priority 0 if user not specify, i don't know what right solution....

my mistake, this is mandatory param, so pointer not needed. i'm fix it

qos.go
```diff
@@ -16,12 +16,188 @@
...
+		Table:     tableLogicalSwitch,
+		Mutations: []interface{}{mutation},
+		Where:     []interface{}{condition},
+	}
+	operations := []libovsdb.Operation{insertOp, mutateOp}
+
+	return &OvnCommand{operations, odbi, make([][]map[string]interface{}, len(operations))}, nil
+}
+
+func (odbi *ovnDBImp) delQoSImp(ls string, direction string, priority *int, match string) (*OvnCommand, error) {
```
softwarejl: why priority is point?  no default value?

vtolstov: without pointer we can't allow wildcard removing of qos rules like ovn-nbctl qos-del allows, so we need pointer in this place.

qos.go
```diff
@@ -16,12 +16,188 @@
...
+
+	odbi.cachemutex.RLock()
+	defer odbi.cachemutex.RUnlock()
+
+	cacheLogicalSwitch, ok := odbi.cache[tableLogicalSwitch]
+	if !ok {
+		return nil, ErrorNotFound
+	}
+
+	for _, drows := range cacheLogicalSwitch {
```
softwarejl: is it possible to have common function for that, so many duplicated code now. 

interface := drows.Fields["xxxx"]
interface.(type)
switch.....

not sure it works.

vtolstov: i'm try to think about this, but now i don't know how to do that.

qos.go
softwarejl: OK.

acl.go
```diff
@@ -42,30 +48,34 @@ func (odbi *ovnDBImp) getACLUUIDByRow(lsw, table string, row OVNRow) (string, er
...
 							if va, ok := a.(libovsdb.UUID); ok {
+								cacheACL, ok := odbi.cache[tableACL][va.GoUUID]
+								if !ok {
+									return "", ErrorNotFound
+								}
 								for field, value := range row {
 									switch field {
 									case "action":
-										if odbi.cache[tableACL][va.GoUUID].Fields["action"].(string) != value {
+										if cacheACL.Fields["action"].(string) != value {
```
softwarejl: in future, maybe have head file to define the table structure, and here can be simplified to (TABLE field name), (TABLE field type).

thoughts?

vtolstov: i'm prefer to try to utilize hashicorp/mapstructure as row have map[string]interface{} format we can automatic unmrashal this row to struct, for types like OvsSet and other specific we can use custom decoder...
So huge part of code can be replaced.

vtolstov:
@softwarejl i have trouble for qos deletion, as i see ovn-nbctl send this request:
```
2019-03-22T17:05:34Z|00024|jsonrpc|DBG|unix:/var/run/openvswitch/ovnnb_db.sock: send request, method="transact", params=["OVN_Northbound",{"rows":[{"qos_rules":["set",[["uuid","199ba4ba-8cc6-44ef-abab-42b94d078650"],["uuid","25695f8d-d1e3-4381-9973-adc841b3a00c"],["uuid","f20a496c-440f-4fa7-abaf-28b7743699dc"]]]}],"until":"==","where":[["_uuid","==",["uuid","83e587e9-00da-42e3-b0a8-631174ae71c8"]]],"columns":["qos_rules"],"timeout":0,"op":"wait","table":"Logical_Switch"},{"where":[["_uuid","==",["uuid","83e587e9-00da-42e3-b0a8-631174ae71c8"]]],"row":{"qos_rules":["set",[]]},"op":"update","table":"Logical_Switch"},{"comment":"ovs-nbctl: ovn-nbctl -v qos-del TEST_LSW to-lport","op":"comment"}], id=4
```

I'm delete 3 qos rules from logical switch.

acl.go
softwarejl: same idea with me,  could we have issue ticket to track it and discuss.

vtolstov: https://github.com/eBay/go-ovn/issues/29 needs more generic name..

vtolstov:
@softwarejl @hzhou8 as i see qos test run without errors now, and api mimics ovn-nbctl.

vtolstov:
i'm rebase and aff comments

PR #35, opened 2019-03-26
Only add load_balancer to lswitches meant to serve LB traffic

noah8713:


load_balancer.go
```diff
@@ -87,24 +89,32 @@ func (odbi *ovnDBImp) lbAddImp(name string, vipPort string, protocol string, add
...
-	mutateOp := libovsdb.Operation{
-		Op:        opMutate,
-		Table:     tableLogicalSwitch,
-		Mutations: []interface{}{mutation},
-		Where:     []interface{}{condition},
+	operations = append(operations, insertOp)
+
+	// Do not mutate all lswitches; instead only for the ones which are meant
+	// to serve LB traffic.
+	LbLswitches := os.Getenv("LB_LSWITCHES")
```
hzhou8: Environment variable is not the right way to pass parameter to API. The API LBAdd should simply add the LB, without attaching to any LS, and we should have a separate API to add/remove a LB to/from LS(es).

vtolstov: yes, we need to follow nbctl
```
lr-lb-add ROUTER LB       add a load-balancer to ROUTER
  lr-lb-del ROUTER [LB]     remove load-balancers from ROUTER
  lr-lb-list ROUTER         print load-balancers
  ls-lb-add SWITCH LB       add a load-balancer to SWITCH
  ls-lb-del SWITCH [LB]     remove load-balancers from SWITCH
  ls-lb-list SWITCH         print load-balancers
```

noah8713: > Environment variable is not the right way to pass parameter to API. The API LBAdd should simply add the LB, without attaching to any LS, and we should have a separate API to add/remove a LB to/from LS(es).

Sure. Done! PTAL at revised PR

noah8713: > yes, we need to follow nbctl
> 
> ```
>   ls-lb-add SWITCH LB       add a load-balancer to SWITCH
>   ls-lb-del SWITCH [LB]     remove load-balancers from SWITCH
>   ls-lb-list SWITCH         print load-balancers
> ```

@vtolstov : Sure we can do that. I added the code in LB instead instead of changing logical_switch.go. Let me know if we still want to prefer nbctl way!

noah8713: @vtolstov : Revised as per nbctl format. PTAL 
CC @hzhou8

load_balancer.go
```diff
@@ -117,36 +101,6 @@ func (odbi *ovnDBImp) lbDelImp(name string) (*OvnCommand, error) {
 		Table: tableLoadBalancer,
 		Where: []interface{}{condition},
 	}
-
-	// Also delete references from Logical switches
```
hzhou8: Please keep this logic. It is the purpose of the PR https://github.com/eBay/go-ovn/pull/33, and it is still correct.

logical_router_lb.go
```diff
@@ -0,0 +1,122 @@
+/**
```
hzhou8: I think one file for operations for a table is enough, no need to add new files for handling each column.
(P.S. router LB support may be added separately, since it is not directly related to this commit.)

ovnnbimp_test.go
```diff
@@ -475,4 +475,73 @@ func TestLoadBalancer(t *testing.T) {
...
+	t.Logf("Adding LB to LRouter %s", LR)
+	ocmd, err = ovndbapi.LRLBAdd(LR, "lb2")
+	if err != nil {
+		t.Fatal(err)
+	}
+	err = ovndbapi.Execute(ocmd)
+	if err != nil {
+		t.Fatalf("Adding LB lb2 to LRouter failed with err %v", err)
+	}
+	t.Logf("Adding LB lb2 to LRouter %s Done", LR)
```
hzhou8: List it to make sure it is added successfully.

ovnnbimp_test.go
```diff
@@ -475,4 +475,73 @@ func TestLoadBalancer(t *testing.T) {
...
+	t.Logf("Deleting LB lb2 from LRouter %s", LR)
+	ocmd, err = ovndbapi.LRLBDel(LR, "lb2")
+	if err != nil {
+		t.Fatal(err)
+	}
+	err = ovndbapi.Execute(ocmd)
+	if err != nil {
+		t.Fatalf("Deleting LB lb2 from LRouter failed with err %v", err)
+	}
+	t.Logf("Deleting LB lb2 from LRouter %s Done", LR)
```
hzhou8: list it to make sure it is deleted successfully

apis.go
```diff
@@ -65,12 +71,20 @@ type OVNDBApi interface {
...
+	// List Load balancers for a LR
+	LRLBList(lr string) ([]*LoadBalancer, error)
 	// Add LB
 	LBAdd(name string, vipPort string, protocol string, addrs []string) (*OvnCommand, error)
 	// Delete LB with given name
 	LBDel(name string) (*OvnCommand, error)
 	// Update existing LB
 	LBUpdate(name string, vipPort string, protocol string, addrs []string) (*OvnCommand, error)
+	// List existing LBs
+	LBList() (*OvnCommand, error)
```
vtolstov: why this return *OvnCommand? as i see before my prs list returns cached data and not create request to ovsdb.

load_balancer.go
```diff
@@ -190,3 +144,17 @@ func (odbi *ovnDBImp) rowToLB(uuid string) (*LoadBalancer, error) {
 
 	return lb, nil
 }
+
+
+
+func (odbi *ovnDBImp) lbListImp() (*OvnCommand, error) {
```
vtolstov: i think that this useless, because we already have such data in cache (or via initial sync, or after lb created via monitor)

logical_router_lb.go
```diff
@@ -0,0 +1,122 @@
...
+	}
+	condition := libovsdb.NewCondition("name", "==", lr)
+	mutateOp := libovsdb.Operation{
+		Op:        opMutate,
+		Table:     tableLogicalRouter,
+		Mutations: []interface{}{mutation},
+		Where:     []interface{}{condition},
+	}
+	operations = append(operations, mutateOp)
+	return &OvnCommand{operations, odbi, make([][]map[string]interface{}, len(operations))}, nil
```
vtolstov: why you don't check passed data? i think that we need to get from cache data for provided lb and lr and return ErrorNotExist when uuid not in cache. @hzhou8 what you think?

logical_switch_lb.go
```diff
@@ -0,0 +1,121 @@
...
+)
+
+
+func (odbi *ovnDBImp) lslbAddImp(lswitch string, lb string) (*OvnCommand, error) {
+	var operations []libovsdb.Operation
+	namedUUID, err := newRowUUID()
+	if err != nil {
+		return nil, err
+	}
+	mutateUUID := []libovsdb.UUID{{namedUUID}}
```
vtolstov: need check provided uuids i think

ovnnbimp_test.go
```diff
@@ -475,4 +475,73 @@ func TestLoadBalancer(t *testing.T) {
 		t.Fatalf("error: lb deletion not done, total:%v", len(lb))
 	}
 	t.Logf("LB deletion done")
+	// test adding lb to lr and lswitch
```
vtolstov: i'm prefer to split this big file to smaller pieces and add tests for specific files, like logical_switch_test.go load_balancer_test.go and so.
I think that this is more idiomatic with other go projects.

apis.go
noah8713: Yeah sorry typo. Forgot to add UT for that. Should have captured.

load_balancer.go
noah8713: ok I thought you wanted LBdel to do LB del only. Will restore

logical_router_lb.go
noah8713: I added because upstream doesnt have any table called Logical_Switch_Load_Balancer and Logical_Router_Load_Balancer. Hence, in lieu of that format, I created new file for load_balancer too. Is

logical_router_lb.go
noah8713: I found one issue in UT. populate cache only populates entries for add and delete. cache doesnt update on mutate ops. Need to fix that too. But I am thinking it in separate commit as this itself is big commit. Is it ok ?

logical_switch_lb.go
noah8713: ok like check if uuid exists?

ovnnbimp_test.go
noah8713: Yeah will do

ovnnbimp_test.go
noah8713: sure. Will do.

ovnnbimp_test.go
noah8713: sure. Will do.

logical_router_lb.go
hzhou8: If you look at the current file structure, each file maps to a table. It is better to follow that. As you said there is no such table like Logical_Switch_Load_Balancer/Logical_Router_Load_Balancer, so you don't need a new file for that. You are operating Logical_Switch/Logical_Router table in this case, so just update those files.

logical_router_lb.go
hzhou8: Agree with @vtolstov validation is needed.

vtolstov: yes, for mutate cache update is nice to have in separate pr

logical_switch_lb.go
vtolstov: yes, i think ok like check for both uuids is fine

load_balancer.go
noah8713: I added this because we dont have any function to list load_balancers.

logical_router_lb.go
noah8713: Cool. Sounds good!

ovnnbimp_test.go
noah8713: @hzhou8 List will not show because we need to add update operations for cache else cache only updates insert and delete data. Do you want me to handle that in same PR? I already have mentioned this to @vtolstov and we can use different PR for updates!

ovnnbimp_test.go
noah8713: same as above!

logical_switch_lb.go
noah8713: Done

logical_router_lb.go
noah8713: Ack!

apis.go
noah8713: Got rid of it!

logical_router.go
```diff
@@ -145,3 +146,104 @@ func (odbi *ovnDBImp) GetLogicalRouters() ([]*LogicalRouter, error) {
...
 }
+
+func (odbi *ovnDBImp) lrlbAddImp(lr string, lb string) (*OvnCommand, error) {
+	var operations []libovsdb.Operation
+	row := make(OVNRow)
+	row["name"] = lb
+	lbuuid := odbi.getRowUUID(tableLoadBalancer, row)
+	if len(lbuuid) == 0 {
+		return nil, ErrorNotFound
+	}
```
vtolstov: Can you add check for lr too?

logical_router.go
```diff
@@ -145,3 +146,104 @@ func (odbi *ovnDBImp) GetLogicalRouters() ([]*LogicalRouter, error) {
...
+	operations = append(operations, mutateOp)
+	return &OvnCommand{operations, odbi, make([][]map[string]interface{}, len(operations))}, nil
+}
+
+func (odbi *ovnDBImp) lrlbDelImp(lr string, lb string) (*OvnCommand, error) {
+	var operations []libovsdb.Operation
+	row := make(OVNRow)
+	row["name"] = lb
+	lbuuid := odbi.getRowUUID(tableLoadBalancer, row)
```
vtolstov: Same check for lr too

ovnnbimp_test.go
hzhou8: Any change to the DB should have been reflected back to the cache already, otherwise, existing tests should fail. For example: https://github.com/eBay/go-ovn/blob/master/dhcp_options_test.go#L101

If there is really a bug, this feature is anyway broken, so merging the PR doesn't help. We should create a new PR to fix the bug and then we can merge this PR. In any case, the test case for this PR should ensure the correctness of this feature, i.e. we should be able to immediately read whatever changes we made to DB.

noah8713: @hzhou8  . The problem was lrouter was getting deleted in the router test cases and hence listing lb from lr was failing when I added more debug. It's ok. Have addressed the same and have added two seperate test cases for lbls and lblr along with assert verification. PTAL.

logical_router.go
noah8713: Done

logical_router.go
noah8713: @vtolstov : Also added for LSwitch too :)

vtolstov: ok

PR #42, opened 2019-03-30
cleanup api implementation

vtolstov:


vtolstov:
@hzhou8 this is first try to cleanup and rename all stuff.
I don't know about idiomatic names for methods like ASAdd...
May be Client need to have methods like
AddressSet()
LogicalSwitch()
...
they returns interface type with needed methods. For ex LogicalSwitch() return 
type LogicalSwitch interface {
Add(xxx)
Del(ssss)
List()
Get(xxx)
}

what do you think?

vtolstov:
this pr waiting for https://github.com/eBay/libovsdb/pull/11

vtolstov:
rebased, merged and fixed

vtolstov:
@hzhou8 please check this too, if all ok and it will be merged i can push first steps to automatic unmarshaling map[string]interface{} to struct of needed type

client.go
```diff
@@ -0,0 +1,370 @@
...
+
+	// Add ACL
+	ACLAdd(ls, direct, match, action string, priority int, external_ids map[string]string, logflag bool, meter string) (*OvnCommand, error)
+	// Delete acl
+	ACLDel(ls, direct, match string, priority int, external_ids map[string]string) (*OvnCommand, error)
+	// Get all acl by lswitch
+	ACLList(ls string) ([]*ACL, error)
+
+	// Get AS
+	ASGet(name string) ([]*AddressSet, error)
```
hzhou8: It doesn't make sense to return a list, because the ovn-nb schema requires that name is unique in the Address_Set table.

vtolstov: yes, fixed

PR #43, opened 2019-04-02
Add external-ids add/del support lswitches

noah8713:


logical_switch.go
```diff
@@ -271,3 +271,57 @@ func (odbi *ovnDBImp) lslblistImp(lswitch string) ([]*LoadBalancer, error) {
...
+
+func (odbi *ovnDBImp) lsExtIdsAddImp(lswitch string, external_ids map[string]string) (*OvnCommand, error) {
+	var operations []libovsdb.Operation
+	row := make(OVNRow)
+	row["name"] = lswitch
+	lsuuid := odbi.getRowUUID(tableLogicalSwitch, row)
+	if len(lsuuid) == 0 {
+		return nil, ErrorNotFound
+	}
+	if external_ids == nil {
```
softwarejl: how about the map is empty?

logical_switch.go
```diff
@@ -271,3 +271,57 @@ func (odbi *ovnDBImp) lslblistImp(lswitch string) ([]*LoadBalancer, error) {
...
+		return nil, fmt.Errorf("No external ids to set.")
+	}
+	mutateSet, err := libovsdb.NewOvsMap(external_ids)
```
softwarejl: if the external_ids is empty, the err = nil, right?

noah8713: if empty, it means we raise error because we are trying to add/del empty ext_id maps

logical_switch.go
noah8713: if empty, it means we raise error because we are trying to add/del empty ext_id maps for a lswitch

noah8713: @softwarejl Handled the empty external_id and just returning in case the its null instead of raising error. Also covered the UT for the same.

logical_switch.go
```diff
@@ -271,3 +271,57 @@ func (odbi *ovnDBImp) lslblistImp(lswitch string) ([]*LoadBalancer, error) {
...
+
+func (odbi *ovnDBImp) lsExtIdsDelImp(lswitch string, external_ids map[string]string) (*OvnCommand, error) {
+	var operations []libovsdb.Operation
+	row := make(OVNRow)
+	row["name"] = lswitch
+	lsuuid := odbi.getRowUUID(tableLogicalSwitch, row)
+	if len(lsuuid) == 0 {
+		return nil, ErrorNotFound
+	}
+	if len(external_ids) == 0 {
```
softwarejl: @hzhou8  is it correct?

vtolstov: i think that adding additional method is overhead. We need to have method that receives struct of kind logical_switch or another, and getRowUUID based on this struct converted to map(already have such method). So we can fill any of fields and get back rows that matched.

logical_switch.go
hzhou8: So you mean we *SHOULD* raise error, but you didn't raise error because you are returning nil, nil at L284. Why?

	if len(external_ids) == 0 {
		return nil, nil
	}

logical_switch.go
hzhou8: @vtolstov I agree in general we should make the interfaces more generic. We will see how your POC works. Before that is ready, we may still need to add methods for the functions we need. So I am ok with the approach of this patch for now.

logical_switch_external_ids_test.go
```diff
@@ -0,0 +1,124 @@
+/**
```
hzhou8: Thanks for adding tests. Could you rename this file to logical_switch_test? I don't think we have one yet.

apis.go
```diff
@@ -58,6 +58,11 @@ type OVNDBApi interface {
 	// List Load balancers for a LSW
 	LSLBList(lswitch string) ([]*LoadBalancer, error)
 
+	// Add ExtIDs to SWITCH
+	LSExtIdsAdd(lswitch string, external_ids map[string]string) (*OvnCommand, error)
```
vtolstov: Why separate method? Why not create LSSet and use it for setting all variables in feature that present in logical switch and not related to weak or strong references to other tables?

logical_switch.go
```diff
@@ -271,3 +271,57 @@ func (odbi *ovnDBImp) lslblistImp(lswitch string) ([]*LoadBalancer, error) {
 	}
 	return listLB, nil
 }
+
+func (odbi *ovnDBImp) lsExtIdsAddImp(lswitch string, external_ids map[string]string) (*OvnCommand, error) {
```
vtolstov: i'm prefer name logical switch with ls and not lswitch, but this is my IMHO

logical_switch.go
```diff
@@ -271,3 +271,57 @@ func (odbi *ovnDBImp) lslblistImp(lswitch string) ([]*LoadBalancer, error) {
...
+	var operations []libovsdb.Operation
+	row := make(OVNRow)
+	row["name"] = lswitch
+	lsuuid := odbi.getRowUUID(tableLogicalSwitch, row)
+	if len(lsuuid) == 0 {
+		return nil, ErrorNotFound
+	}
+	if len(external_ids) == 0 {
```
vtolstov: i'm prefer to check like external_ids == nil ?
i don't think that somebody pass non nil empty map.

apis.go
vtolstov: @hzhou8 what you think?

logical_switch.go
hzhou8: I think validation is needed for both. If external_ids == nil || len(external_ids) == 0

logical_switch.go
```diff
@@ -271,3 +271,57 @@ func (odbi *ovnDBImp) lslblistImp(lswitch string) ([]*LoadBalancer, error) {
...
+func (odbi *ovnDBImp) lsExtIdsAddImp(ls string, external_ids map[string]string) (*OvnCommand, error) {
+	var operations []libovsdb.Operation
+	row := make(OVNRow)
+	row["name"] = ls
+	lsuuid := odbi.getRowUUID(tableLogicalSwitch, row)
+	if len(lsuuid) == 0 {
+		return nil, ErrorNotFound
+	}
+	if len(external_ids) == 0 {
+		return nil, fmt.Errorf("external_ids is empty. Please provide a valid key-value pair to update lswitch")
```
hzhou8: The error can be simply "external_ids is nil or empty"

logical_switch.go
hzhou8: Great. Thanks for explain. I am new to Go :)

noah8713:
@vtolstov Please approve and merge the same! Got one LGTM from Han

PR #44, opened 2019-04-16
WIP: test rewrite

vtolstov:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 language: go
 
 go:
+  - 1.11.x
   - 1.12.x
 
 os:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 language: go
 
 go:
-  - 1.11.x
   - 1.12.x
 
 os:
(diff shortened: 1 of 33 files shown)


hzhou8:
Hi @vtolstov, this looks promising but you didn't update it for a while. Could you share your plan on this? Thanks!

vtolstov:
Hi and sorry, busy at other tasks. I'm plan to complete this in this week.

vtolstov:
@hzhou8 @lebauce can you look at acl.go in this pr.
I think that it looks better with some reflection based stuff in ovnnbimp.go. Also for listing we don't need reconstruct item each time and save allocations.
But i don't like getRows in ovnnbimp.go (i think that it can looks simpler then now)

vtolstov:
@hzhou8 may be i found the solution.
goovn store all data in cache with map[string]interface{}
when user wants some struct - convert from map to native struct
when user wants to add some object (acl, qos, logical switch...) he don't provide map, but use something like this
```
type Options map[string]interface{}
ACLAdd(target string, opts ...ACLOpt) cmd, error
ACLOpt is the type Option func(Options)

for example for ACLAdd options is like:
func Direction(d string) Option {
  return func(o Options) {
    o["direction"] = d
  }
}

func Priority(p int) Option {
  return func(o Options) {
    o["priority"] = p
  }
}

func Match(m string) Option {
  return func(o Options) {
    o["match"] = m
  }
}

func Action(a string) Option {
  return func(o Options) {
    o["action"] = a
  }
}
```

vtolstov:
so we always know what params we need to pass, and have ability to validate it.
to check existence of item we can iterate over map (this is simpler when my previous reflection based stuff)

vtolstov:
abandon

PR #46, opened 2019-06-17
Fix connect string for ssl

noah8713:
Current connect string to libovsdb always passes tcp.
Hence, ssl will not work if client wants to use ssl.
This commit fixes the same.


softwarejl:
LGTM

PR #48, opened 2019-07-12
show lb , ports ,sr in LRGet

JackyCZJ:
Find it can't show any LoadBlancer , ports, staticRouter information in LRGet ,try to add it.

### DIFF
--- a/logical_router.go
+++ b/logical_router.go
@@ -118,13 +118,74 @@ func (odbi *ovndb) rowToLogicalRouter(uuid string) *LogicalRouter {
 		}
 	}
 
-	ports := odbi.cache[tableLogicalRouter][uuid].Fields["ports"]
-	switch ports.(type) {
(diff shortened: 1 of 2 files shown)


vtolstov:
can you add test to such options? For example attach LoadBalancer to LogicalRouter and after that get LogicalRouter and check that LoadBalancer column contains needed data?

JackyCZJ:
, I just recived right data form my option, I will add some test tomorrow . Sorry for poor English.

vtolstov:
thanks, i think that you do something wrong. new commits not goes to this pr.
Can you fix it?

JackyCZJ:
Yes , actually I did something wrong so I reset it , I will fix it at tomorrow cuz  I can't code nomally right now, Sorry for border you.

vtolstov:
thanks!

PR #51, opened 2019-07-19
Process table updates in the right order when populating cache

lebauce:
When a logical router update (https://github.com/eBay/go-ovn/blob/84e1e53ffe4fbd63d1b1b4ec3124fae5fb9827cb/ovnnbimp.go#L205) is processed, the `ports` field is considered. If the logical router ports were not processed before, the call to `rowToLogicalRouterPort` (https://github.com/eBay/go-ovn/blob/84e1e53ffe4fbd63d1b1b4ec3124fae5fb9827cb/logical_router.go#L154) will crash as the ports are missing from the cache.

This is because the table updates is a `map[string]TableUpdate`, so there is no guarantee on the tables processing order.

### DIFF
--- a/common.go
+++ b/common.go
@@ -49,3 +49,25 @@ const (
 	tableSSL                      string = "SSL"
 	tableGatewayChassis           string = "Gateway_Chassis"
 )
+
+var tablesOrder = []string{
+	tableNBGlobal,
+	tableLogicalSwitchPort,
+	tableAddressSet,
+	tablePortGroup,
+	tableACL,
+	tableQoS,
+	tableMeter,
+	tableMeterBand,
+	tableLogicalRouterPort,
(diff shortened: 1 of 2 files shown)


JackyCZJ:
So there may  have three carsh at this type ? I write this for show  information.  Sorry for make such mistake if it causing failure.
https://github.com/eBay/go-ovn/blob/84e1e53ffe4fbd63d1b1b4ec3124fae5fb9827cb/logical_router.go#L129
https://github.com/eBay/go-ovn/blob/84e1e53ffe4fbd63d1b1b4ec3124fae5fb9827cb/logical_router.go#L154
https://github.com/eBay/go-ovn/blob/84e1e53ffe4fbd63d1b1b4ec3124fae5fb9827cb/logical_router.go#L176

PR #54, opened 2019-07-24
fix goreportcard issues

vtolstov:


vtolstov:
@lebauce may the current travis tests failed because of some changes to deleting items in cache?
(i'm try to fix some issues founded with goreportcard and found this)

lebauce:
@vtolstov strange. I'm checking

lebauce:
@vtolstov It seems the `cmd, err = ovndbapi.ACLAdd(LSW, "to-lport", MATCH, "drop", 1001, nil, true, "")` is done twice in acl_test.go Therefore, the check you added to the second one will fail, the test will exit leaving the logical switch it created, making the following tests fail too. We should probably use `defer` to delete the objects created during the test.

lebauce:
@vtolstov exactly

PR #56, opened 2019-07-25
Fix DHCP Option set

JackyCZJ:


hzhou8:
@JackyCZJ Thanks for the patch! It would be helpful if you could add commit message to describe briefly what's the problem and how is it solved.

JackyCZJ:
Like :" Fixd: Use uuid instead of cidr as index to set DHCP Option . "?  my poor English lol

hzhou8:
@JackyCZJ it is better. But in general, if the title doesn't provide enough details, we should add more information in the commit message body. If you are fixing something, briefly describe the problem so that reviewer quickly knows the purpose. Briefly describe the solution would also be helpful if the solution to the problem is not obvious.

Here is an example (maybe not perfect, but you can see what I meant):

commit aa10a9506bb5e4abf03508c2ffabd09b05868024
Author: Han Zhou <hzhou8@ebay.com>
Date:   Fri Mar 8 11:52:01 2019 -0800

    Fix test init.
    
    OVS sandbox NB DB socket name changed from ovn-nb.sock to nb1.ovsdb,
    which causes existing test initialization failing. This patch makes
    it smarter to read from $OVN_NB_DB env.

hzhou8:
@vtolstov added the DHCP APIs initially. Need him to confirm why not using uuid, or any concern about using uuid.

dhcp_options.go
```diff
@@ -91,26 +91,43 @@ func (odbi *ovndb) dhcpOptionsAddImp(cidr string, options map[string]string, ext
...
+		if err != nil {
+			return nil, err
+		}
+		row["external_ids"] = oMap
+	}
 	condition := libovsdb.NewCondition("_uuid", "==", libovsdb.UUID{dhcpUUID})
 
 	mutateOp := libovsdb.Operation{
-		Op:        opMutate,
+		Op:        opUpdate,
```
hzhou8: The original purpose of this API (I think) is to add option key-value pairs for existing cidr (and external_ids). This patch is changing it to a complete replacement of all options and external_ids. Could you explain the motivation first?

vtolstov: if we follow ovn-nbctl commands, we don't eed to replace, but append absent, and change exiting options.

vtolstov:
my mistake that initially i think that cidr is uniq, but later i saw that i can have many options created in ovnnb and assign needed options to port via uuid. So we don't need to refer options via cidr, but via uniq uuid of such options.

dhcp_options.go
```diff
@@ -91,26 +91,43 @@ func (odbi *ovndb) dhcpOptionsAddImp(cidr string, options map[string]string, ext
...
 
-func (odbi *ovndb) dhcpOptionsSetImp(cidr string, options map[string]string, external_ids map[string]string) (*OvnCommand, error) {
-
-	row, err := newDHCPRow(cidr, nil, external_ids)
-	if err != nil {
-		return nil, err
+func (odbi *ovndb) dhcpOptionsSetImp(uuid string, options map[string]string, external_ids map[string]string) (*OvnCommand, error) {
+	row := make(OVNRow)
+	dhcpUUID := ""
+	dhcpuuidlist,err := odbi.DHCPOptionsList()
```
vtolstov: why? you need to check only that this uuid exists, so don't over allocate slice for all options.
you already have uuid, so check for ok in odbi.cache[tableDHCPOptions][UUID]

dhcp_options.go
```diff
@@ -91,26 +91,43 @@ func (odbi *ovndb) dhcpOptionsAddImp(cidr string, options map[string]string, ext
...
+	if err != nil{
+		return nil, ErrorNotFound
 	}
-
-	dhcpUUID := odbi.getRowUUID(tableDHCPOptions, row)
-	if len(dhcpUUID) == 0 {
+	for _ ,v := range dhcpuuidlist{
```
vtolstov: this is not needed

dhcp_options.go
```diff
@@ -91,26 +91,43 @@ func (odbi *ovndb) dhcpOptionsAddImp(cidr string, options map[string]string, ext
...
+		if v.UUID == uuid{
+			dhcpUUID = uuid
+			break
+		}
+	}
+	if dhcpUUID == ""{
 		return nil, ErrorNotFound
 	}
+	if options != nil {
```
vtolstov: may be if options is nil - return error?

dhcp_options.go
JackyCZJ: got it , wip

JackyCZJ:
```
root@esixCell:/home/jacky# ovn-nbctl dhcp-options-set-options ac5a44a0-8c38-4f3b-a861-1e1780d6287a "server_id"="192.168.1.1" "server_mac"="54:54:54:54:54:54" "lease_time"="6000"
root@esixCell:/home/jacky# ovn-nbctl list  DHCP_Options
_uuid               : ac5a44a0-8c38-4f3b-a861-1e1780d6287a
cidr                : "192.168.0.0/24"
external_ids        : {}
options             : {lease_time="6000", server_id="192.168.1.1", server_mac="54:54:54:54:54:54"}
root@esixCell:/home/jacky# ovn-nbctl dhcp-options-set-options ac5a44a0-8c38-4f3b-a861-1e1780d6287a "server_id"="192.168.1.1"  "lease_time"="5000"
root@esixCell:/home/jacky# ovn-nbctl list  DHCP_Options
_uuid               : ac5a44a0-8c38-4f3b-a861-1e1780d6287a
cidr                : "192.168.0.0/24"
external_ids        : {}
options             : {lease_time="5000", server_id="192.168.1.1"}
```
this is ovn-nbctl  dhcp-options-set-options , it just completely replaced . To replace like key-value , we need to contrast change to update, becase options all in one field.

dhcp_options.go
JackyCZJ: maybe we can make options to "" to clean all options.....

JackyCZJ:
> @JackyCZJ it is better. But in general, if the title doesn't provide enough details, we should add more information in the commit message body. If you are fixing something, briefly describe the problem so that reviewer quickly knows the purpose. Briefly describe the solution would also be helpful if the solution to the problem is not obvious.
> 
> Here is an example (maybe not perfect, but you can see what I meant):
> 
> commit [aa10a95](https://github.com/eBay/go-ovn/commit/aa10a9506bb5e4abf03508c2ffabd09b05868024)
> Author: Han Zhou [hzhou8@ebay.com](mailto:hzhou8@ebay.com)
> Date: Fri Mar 8 11:52:01 2019 -0800
> 
> ```
> Fix test init.
> 
> OVS sandbox NB DB socket name changed from ovn-nb.sock to nb1.ovsdb,
> which causes existing test initialization failing. This patch makes
> it smarter to read from $OVN_NB_DB env.
> ```

thanks for suggestion !

dhcp_options.go
JackyCZJ: done

dhcp_options.go
JackyCZJ: done

dhcp_options.go
JackyCZJ: donw

dhcp_options.go
vtolstov: Ok if ovn-nbctl replaces options do the same

JackyCZJ: wait , i commit my test env .... I should fix it

JackyCZJ:
@hzhou8 @vtolstov PTAL

client.go
```diff
@@ -388,20 +390,24 @@ func (c *ovndb) DHCPOptionsAdd(cidr string, options map[string]string, external_
...
+	return c.dhcpOptionsGetImp(uuid)
+}
+
 func (c *ovndb) DHCPOptionsList() ([]*DHCPOptions, error) {
 	return c.dhcpOptionsListImp()
 }
 
 func (c *ovndb) LRNATAdd(lr string, ntype string, externalIp string, logicalIp string, external_ids map[string]string, logicalPortAndExternalMac ...string) (*OvnCommand, error) {
-	return c.lrNatAddImp(lr, ntype, externalIp, logicalIp,external_ids, logicalPortAndExternalMac...)
+	return c.lrNatAddImp(lr, ntype, externalIp, logicalIp, external_ids, logicalPortAndExternalMac...)
```
vtolstov: unrelated to this pr change, please use goimports with you code editor.

dhcp_options.go
```diff
@@ -91,27 +91,37 @@ func (odbi *ovndb) dhcpOptionsAddImp(cidr string, options map[string]string, ext
...
+func (odbi *ovndb) dhcpOptionsSetImp(uuid string, options map[string]string, external_ids map[string]string) (*OvnCommand, error) {
+	row := make(OVNRow)
 
-	row, err := newDHCPRow(cidr, nil, external_ids)
+	_, ok := odbi.cache[tableDHCPOptions][uuid]
+	if !ok {
+		return nil, ErrorNotFound
+	}
+	if options == nil {
+		return nil, ErrorSchema
```
vtolstov: use different error type for invalid args - ErrorOption

vtolstov:
i'm fix minor suggestions and push

PR #59, opened 2019-07-29
test

vtolstov:


vtolstov:
please, take a look at client.go and logical_switch.go
I think that this is good enough for new api. It based on new pr in libovsdb.

Main benefits - Client Have dedicated interfaces for each OVN Row. LogicalSwitch, LoadBalancer, ACL.
On each interface we have simple methods like Add, Del, List, Get, Set...
Each method have variable options... for later dedicated stuff we can create options for each method, like LogicalSwitchAddOpt, so we can pass to Add method only needed options.

JackyCZJ:
when will you apply this patch ? it look much better than now ... maybe you can make an new branch to process it ?

vtolstov:
Thanks. You can help by translating other missing pieces like qos, dhcp to never syntax.

vtolstov:
close, as all stuff must be handled by libovsdb

PR #64, opened 2019-08-19
Fix Unit tests:

noah8713:
Unit tests always work fine on first time. Hence, travis build always passes.
Re-running same tests againts same ovn setup fails as entities that were
created failed to clean up.
Hence, this patch fixes the same by handling all such cases.

### DIFF
--- a/acl_test.go
+++ b/acl_test.go
@@ -71,11 +71,12 @@ func TestACLs(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	lsws, err := ovndbapi.LSList()
+	lsws, err := ovndbapi.LSGet(LSW)
 	if err != nil {
 		t.Fatal(err)
 	}
-	if len(lsws) != 1 {
+
+	if len(lsws) == 0 {
 		t.Fatalf("ls not created %d", len(lsws))
 	}
 
--- a/dhcp_options_test.go
+++ b/dhcp_options_test.go
@@ -23,12 +23,17 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
+const (
+	LSW2 = "TEST_LSW2"
+	LSP2 = "TEST_LSP"
+)
+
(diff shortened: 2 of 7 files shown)


acl_test.go
```diff
@@ -75,7 +75,8 @@ func TestACLs(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	if len(lsws) != 1 {
+
+	if len(lsws) == 0 {
```
vtolstov: i don't think that this is good solution. for example - you run second test and lsw not created, but you have older lsw....
best to do - count number of lsw before creating and after.

dhcp_options_test.go
```diff
@@ -96,7 +101,7 @@ func TestDHCPOptions(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	if len(lsws) != 1 {
+	if len(lsws) == 0 {
```
vtolstov: same as before

acl_test.go
noah8713: count logic itself needs refactoring.  Right now we are using multiple LSW and hence hardcoding to ==1 doesn't make sense. Hence, I kept 0. To avoid check by length, we should get list of switches and then iterate by ls.Name==lsw to do actual confirmation if the switch is deleted/created. Similar is needed for all entities. Hence, I didnt handle it in this patch. Let me know and I can amend in same patch.

dhcp_options_test.go
noah8713: same as above.

acl_test.go
noah8713: @vtolstov : Done. PTAL.

dhcp_options_test.go
noah8713: @vtolstov : Done. PTAL.

softwarejl:
LGTM

vtolstov:
thank you

PR #65, opened 2019-10-16
update libovsdb version to v0.2.0

jxiaobin:
a new tag v0.2.0 has been created in libovsdb project, updating go-ovn
accordingly for better semantic version management.

### DIFF
--- a/.travis/test.sh
+++ b/.travis/test.sh
@@ -1,7 +1,7 @@
 #!/bin/sh -ex
 
 rm -rf ovs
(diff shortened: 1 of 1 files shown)


hzhou8:
We need to fix the Travis before merging. Also, @jxiaobin it would be better if you could describe the reason of this change, i.e. what was broken without this change. Thanks!

jxiaobin:
Travis failed because upstream ovs/ovn is broken.

jxiaobin:
switched ovs to branch-2.12 branch, travis passed

PR #66, opened 2019-11-21
pull ovs/branch-2.12 to test

jxiaobin:
due to ovs/ovn split, ovs master doesn't work with go-ovn test,
use branch-2.12 instead.


jxiaobin:
@hzhou8 plz take a look.

vtolstov:
as this is trivial change to test script, i'll merge it

PR #67, opened 2019-11-25
add function to link switch to router

jxiaobin:


jxiaobin:
@vtolstov @hzhou8 could you help review? thanks!

JackyCZJ:
Maybe you can add some test for your change.

client.go
```diff
@@ -39,6 +39,8 @@ type Client interface {
 
 	// Add logical port PORT on SWITCH
 	LSPAdd(ls string, lsp string) (*OvnCommand, error)
+	// Link logical port to router
+	LSPLinkToRouter(ls string, lsp, lrp string) (*OvnCommand, error)
```
hzhou8: It is creating a new LSP and link to an existed LRP, so the API is better to be named as LSPAddToLRP.

logical_switch_port.go
```diff
@@ -73,6 +73,58 @@ func (odbi *ovndb) lspAddImp(lsw, lsp string) (*OvnCommand, error) {
...
+
+	mutateOp := libovsdb.Operation{
+		Op:        opMutate,
+		Table:     tableLogicalSwitch,
+		Mutations: []interface{}{mutation},
+		Where:     []interface{}{condition},
+	}
+
+	options := make(map[string]string)
+	options["router-port"] = lrp
```
hzhou8: Would it be better to validate if the LRP exists?

logical_switch_port.go
```diff
@@ -73,6 +73,58 @@ func (odbi *ovndb) lspAddImp(lsw, lsp string) (*OvnCommand, error) {
...
+	optMap, _ := libovsdb.NewOvsMap(options)
+	// set options
+	setOptionsOp := libovsdb.Operation{
```
hzhou8: Can this be simpler? Can we avoid this setOptionsOp by directly: port["options"] = options, as part of the insertOp?

jxiaobin:
closing this pr, will submit another one to link switch to router in a transaction

Issue #18, opened 2019-03-12
Support multiple API calls in one transaction


hzhou8:

vtolstov:
I think that when we pass multiple commands it runs in single transaction...

hzhou8:
Yes, this is the desired behavior, but it is not supported yet.

PR #68, opened 2019-11-26
Add function to link switch to router

jxiaobin:
The function accepts switch name, switch port name, router name, router port
name, router port mac, network and etc, then does following:
1. create router port
2. create switch port
3. add router port to router
4. add switch port to switch

so that the switch will be connected to router.

### DIFF
--- a/client.go
+++ b/client.go
@@ -36,6 +36,8 @@ type Client interface {
 	LSExtIdsAdd(ls string, external_ids map[string]string) (*OvnCommand, error)
 	// Del external_ids from logical_switch
 	LSExtIdsDel(ls string, external_ids map[string]string) (*OvnCommand, error)
+	// Link logical switch to router
+	LinkSwitchToRouter(lsw, lsp, lr, lrp, lrpMac string, networks []string, externalIds map[string]string) (*OvnCommand, error)
 
 	// Add logical port PORT on SWITCH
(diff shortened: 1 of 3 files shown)


vtolstov:
i'm really need to complete own pr to be able to create free form rows in ovnnb tables

jxiaobin:
If the operation is split into multiple calls, client needs to handle rollback stuff. The purpose of this function is to make the operation transactional, so we don’t have to handle failures in the middle.

vtolstov:
thanks, can we run transaction before changes, add changes and commit transaction?

jxiaobin:
sorry, I didn't get your point, do you mean wrap the transaction in something like begin, operations, then commit blocks like sql?

vtolstov:
yes, as i know ovsdb have transactions support, so may be best way to write code that uses transactions and avoid duplicate logic inside go-ovn functions?

hzhou8:
@vtolstov I agree with you but the feature of combining APIs calls in a single transaction is not ready yet in go-ovn/libovsdb. See https://github.com/eBay/go-ovn/issues/18

So I'd say let's go ahead with @jxiaobin 's approach for now, until we have the complete transaction support ready. What do you think?

vtolstov:
@hzhou8 yes, i'm agree with you.

PR #69, opened 2020-02-09
LS Unit tests fix

noah8713:
commit f356370 added tests for linking lr to ls.
However, it failed to cleanup the LR and LS. Re-running the test
failed. Hence, this patch cleans up entities that failed to cleanup.


vtolstov:
Thanks

PR #70, opened 2020-02-10
Add support for OVN SouthBound DB access.

noah8713:
1. Add chassis add/get/del support
2. Refactor existing code to support access for both OVN NB and
   SB DB.
3. When running ovn nb db tests, ignore ovn sb db tests and vice versa
4. Update ci scripts to run both ovn nb and sb db tests
5. go fmt for existing code

Signed-off-by: Aliasgar Ginwala <aginwala@ebay.com>


vtolstov:
thanks, for feature gofmt of exiting code prefer in separate commit

hzhou8:
@vtolstov Thanks for reviewing PRs. I just want to kindly remind that we require at least two reviewers approve PR before merging it. Thanks for understanding. - Han

vtolstov:
Sorry, I'm forgot. No problem. I'm try to be better in feature

apis.go
```diff
@@ -75,9 +75,12 @@ type OVNSignal interface {
 
 	onMeterBandCreate(band *MeterBand)
 	onMeterBandDelete(band *MeterBand)
+
+	onChassisCreate(ch *Chassis)
```
hzhou8: Better to add a comment stating this is for SB-DB

chassis.go
```diff
@@ -0,0 +1,145 @@
...
+	Encaps              []string
+	ExternalID          map[interface{}]interface{}
+	Hostname            string
+	Name                string
+	NbCfg               int
+	TransportZones      []string
+	VtepLogicalSwitches []string
+}
+
+func (odbi *ovndb) chassisAddImp(name string, hostname string, etype string, ip string) (*OvnCommand, error) {
```
hzhou8: One chassis may have N encaps. The etype and ip should be list type.

chassis_test.go
```diff
@@ -0,0 +1,76 @@
...
+	CHASSIS_HOSTNAME = "fakehost"
+	CHASSIS_NAME     = "fakechassis"
+	ENCAP_TYPE       = "geneve"
+	IP               = "10.0.0.11"
+)
+
+func TestChassis(t *testing.T) {
+	db = getOVNDB()
+	if db == dbNB {
+		t.Skip("Skip running chassis test against nb db")
```
hzhou8: In this case we should check if the API call returns proper error, e.g. "the API belongs to xx-DB but the client is connected to yy-DB"

client.go
```diff
@@ -17,12 +17,14 @@
...
 
 import (
-	"sync"
-
 	"github.com/ebay/libovsdb"
+	"sync"
 )
 
-// Client ovnnb client
+var db string
```
hzhou8: Sorry I didn't see where is this used. I think it should be assigned the DB name that the client is connected to, and then it can be used in each API implementation to ensure the API is called with the correct DB and raise error if it is a mismatch.

client.go
```diff
@@ -173,7 +182,7 @@ type ovndb struct {
 	disconnectCB OVNDisconnectedCallback
 }
 
-func NewClient(cfg *Config) (Client, error) {
+func NewClient(cfg *Config, db string) (Client, error) {
```
hzhou8: This need to be backward compatible. Without specifying , it should by default use NB-DB.

client.go
```diff
@@ -186,7 +195,7 @@ func NewClient(cfg *Config) (Client, error) {
 	}
 	imp.client = c
 
-	initial, err := imp.client.MonitorAll(dbNB, "")
+	initial, err := imp.client.MonitorAll(db, "")
```
hzhou8: "db" should be validated. We should only accept string "OVN_Northbound" and "OVN_Southbound"

ovn_connect_test.go
```diff
@@ -45,32 +46,50 @@ const (
...
 
-var ovndbapi Client
+var (
+	ovndbapi   Client
+	ovn_db     string
+	ovn_socket string
+)
+
+// Default connect to ovn nb db
+func getOVNDB() (db string) {
```
hzhou8: Sorry I don't understand how this works. Usually we will set both OVN_NB_DB and OVN_SB_DB envs to point to corresponding URLs. If both are set, this always returns dbSB? This seems strange.

chassis.go
noah8713: schema shows encap type as string
Encap": {
            "columns": {
                "type": {"type": {"key": {
                           "type": "string",
                           "enum": ["set", ["geneve", "stt", "vxlan"]]}}},

chassis_test.go
vtolstov: yes, we need to distinglish

chassis.go
vtolstov: but via cli app i can set two encaps

noah8713: @vtolstov : Please look at the latest go-ovn master code. This is already addressed  and implemented :)  e.g. ovn-sbctl chassis-add ch1 stt,geneve,vxlan 1.2.3.5

vtolstov: Thanks

PR #71, opened 2020-02-20
SB db chassis and encaps enhancements:

noah8713:
Current code only supports one encap for a chassis.
Hence, add support for adding one or many encaps for the chassis.

e.g. ovn-sbctl chassis-add ch1 stt,geneve,vxlan 1.2.3.5

### DIFF
--- a/chassis.go
+++ b/chassis.go
@@ -17,6 +17,7 @@
 package goovn
 
 import (
+	"fmt"
 	"github.com/ebay/libovsdb"
 )
 
@@ -32,66 +33,79 @@ type Chassis struct {
 	VtepLogicalSwitches []string
 }
 
-func (odbi *ovndb) chassisAddImp(name string, hostname string, etype string, ip string) (*OvnCommand, error) {
-	// / Prepare for encap record
-	enCapUUID, err := newRowUUID()
-	if err != nil {
-		return nil, err
+func (odbi *ovndb) chassisAddImp(name string, hostname string, etype []string, ip string,
+	external_ids map[string]string, transport_zones []string, vtep_lswitches []string) (*OvnCommand, error) {
+	if len(name) == 0 {
+		return nil, fmt.Errorf("chassis name cannot be empty")
 	}
-	row := make(OVNRow)
-
-	if len(name) > 0 {
+	if len(etype) == 0 {
+		return nil, fmt.Errorf("chassis encap type cannot be empty")
+	}
+	if len(ip) == 0 {
+		return nil, fmt.Errorf("chassis ip cannot be empty")
+	}
+	//Prepare for encap record
+	var encap_ids [][]string
+	var namedUUID = "named-uuid"
+	var operations []libovsdb.Operation
+	for _, et := range etype {
+		enCapUUID, err := newRowUUID()
+		if err != nil {
+			return nil, err
+		}
+		row := make(OVNRow)
 		row["chassis_name"] = name
-	}
-
-	if len(etype) > 0 {
-		row["type"] = etype
-	}
-
-	if len(ip) > 0 {
+		var encap_id []string
+		encap_id = append(encap_id, namedUUID)
+		encap_id = append(encap_id, enCapUUID)
+		encap_ids = append(encap_ids, encap_id)
 		row["ip"] = ip
-	}
-	if uuid := odbi.getRowUUID(tableEncap, row); len(uuid) > 0 {
-		return nil, ErrorExist
+		row["type"] = et
+		if uuid := odbi.getRowUUID(tableEncap, row); len(uuid) > 0 {
+			return nil, ErrorExist
+		}
+		insertEncapOp := libovsdb.Operation{
+			Op:       opInsert,
+			Table:    tableEncap,
+			Row:      row,
+			UUIDName: enCapUUID,
+		}
+		operations = append(operations, insertEncapOp)
 	}
 
-	insertEncapOp := libovsdb.Operation{
-		Op:       opInsert,
-		Table:    tableEncap,
-		Row:      row,
-		UUIDName: enCapUUID,
-	}
 	// Prepare for chassis record
 	ChassisUUID, err := newRowUUID()
 	if err != nil {
 		return nil, err
 	}
-
 	rowChassis := make(OVNRow)
-
-	var encap_id []string
-	var namedUUID = "named-uuid"
-	encap_id = append(encap_id, namedUUID)
-	encap_id = append(encap_id, enCapUUID)
-
-	rowChassis["encaps"] = encap_id
-	if len(name) > 0 {
-		rowChassis["name"] = name
+	// consolidate encaps
+	var encaps []interface{}
+	encaps = append(encaps, "set")
+	encaps = append(encaps, encap_ids)
+	rowChassis["encaps"] = encaps
+	rowChassis["name"] = name
+	rowChassis["hostname"] = hostname
+	if external_ids != nil {
+		oMap, err := libovsdb.NewOvsMap(external_ids)
+		if err != nil {
+			return nil, err
+		}
+		rowChassis["external_ids"] = oMap
 	}
-	if len(hostname) > 0 {
-		rowChassis["hostname"] = hostname
+	if len(transport_zones) != 0 {
+		rowChassis["transport_zones"] = transport_zones
 	}
-	if uuid := odbi.getRowUUID(tableChassis, row); len(uuid) > 0 {
-		return nil, ErrorExist
+	if len(vtep_lswitches) != 0 {
+		rowChassis["vtep_logical_switches"] = vtep_lswitches
 	}
-
 	insertChassisOp := libovsdb.Operation{
 		Op:       opInsert,
 		Table:    tableChassis,
 		Row:      rowChassis,
 		UUIDName: ChassisUUID,
 	}
-	operations := []libovsdb.Operation{insertEncapOp, insertChassisOp}
+	operations = append(operations, insertChassisOp)
 	return &OvnCommand{operations, odbi, make([][]map[string]interface{}, len(operations))}, nil
 
 }
@@ -109,7 +123,7 @@ func (odbi *ovndb) chassisDelImp(name string) (*OvnCommand, error) {
 	return &OvnCommand{operations, odbi, make([][]map[string]interface{}, len(operations))}, nil
 }
 
-func (odbi *ovndb) chassisGetImp(hostname string) ([]*Chassis, error) {
+func (odbi *ovndb) chassisGetImp(chassis string) ([]*Chassis, error) {
 	var listChassis []*Chassis
 
 	odbi.cachemutex.RLock()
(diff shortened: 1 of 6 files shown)


noah8713:
@hzhou8 : PTAL

chassis.go
```diff
@@ -32,66 +32,71 @@ type Chassis struct {
...
+				return nil, ErrorExist
+			}
+			insertEncapOp := libovsdb.Operation{
+				Op:       opInsert,
+				Table:    tableEncap,
+				Row:      row,
+				UUIDName: enCapUUID,
+			}
+			// check if chassis already exists with that encap
+			if uuid := odbi.getRowUUID(tableChassis, row); len(uuid) > 0 {
```
hzhou8: How could a row of Encap exists in Chassis table? This whole "if" should be removed.

chassis.go
```diff
@@ -32,66 +32,71 @@ type Chassis struct {
...
-	encap_id = append(encap_id, namedUUID)
-	encap_id = append(encap_id, enCapUUID)
-
-	rowChassis["encaps"] = encap_id
+	// consolidate encaps
+	var encaps []interface{}
+	encaps = append(encaps, "set" )
+	encaps = append(encaps, encap_ids)
+	rowChassis["encaps"] = encaps
 	if len(name) > 0 {
```
hzhou8: "name" is mandatory column for Chassis. It should error out if len(name) == 0.

chassis_test.go
```diff
@@ -23,17 +23,19 @@ import (
...
+var ENCAP_TYPES = []string{ "stt", "geneve", "vxlan" }
+
 func TestChassis(t *testing.T) {
 	db = getOVNDB()
 	if db == dbNB {
 		t.Skip("Skip running chassis test against nb db")
 	}
 	t.Logf("Adding Chassis to OVN SB DB")
-	ocmd, err := ovndbapi.ChassisAdd(CHASSIS_NAME, CHASSIS_HOSTNAME, ENCAP_TYPE, IP)
+	ocmd, err := ovndbapi.ChassisAdd(CHASSIS_NAME, CHASSIS_HOSTNAME, ENCAP_TYPES, IP)
```
hzhou8: Please test for hostname == nil or ""

chassis_test.go
```diff
@@ -23,17 +23,19 @@ import (
...
 	if err != nil {
 		t.Fatal(err)
 	}
```
hzhou8: Please ensure Encaps are written to DB and can be read back correctly. I didn't see implementation of populating encaps back, so there is no way for the current implementation to read encaps, and other fields such as external_ids, transport_zones, etc. of Chassis record.

chassis.go
```diff
@@ -32,66 +32,71 @@ type Chassis struct {
...
-	if err != nil {
-		return nil, err
-	}
-	row := make(OVNRow)
-
-	if len(name) > 0 {
-		row["chassis_name"] = name
-	}
-
+func (odbi *ovndb) chassisAddImp(name string, hostname string, etype []string, ip string) (*OvnCommand, error) {
```
hzhou8: In addition, it is better to support external_ids in the API.

noah8713: Ok I will add

chassis.go
```diff
@@ -32,66 +32,71 @@ type Chassis struct {
...
+			}
+			row := make(OVNRow)
 
-	if len(ip) > 0 {
-		row["ip"] = ip
-	}
-	if uuid := odbi.getRowUUID(tableEncap, row); len(uuid) > 0 {
-		return nil, ErrorExist
+			if len(name) > 0 {
+				row["chassis_name"] = name
```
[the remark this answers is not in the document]
noah8713: I added validations for all mandatory args

chassis.go
```diff
@@ -32,66 +32,71 @@ type Chassis struct {
...
+			}
+			var encap_id []string
+			encap_id = append(encap_id, namedUUID)
+			encap_id = append(encap_id, enCapUUID)
+			encap_ids = append(encap_ids, encap_id)
+			if len(ip) > 0 {
```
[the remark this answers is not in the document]
noah8713: I added validations for all mandatory args

chassis.go
noah8713: My bad as this is mistake while refactor as this condition will never be true, UT didn't catch it. I will move this  out out encap loop and basically its checking if raw with uuid already exists in chassis table before creating a new one

chassis.go
noah8713: I added validations for all mandatory args

chassis_test.go
noah8713: Will do

chassis_test.go
noah8713: Sure will add chassis list and encap list support in revised patch.

chassis.go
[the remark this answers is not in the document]
hzhou8: This comment is about that assigning "chassis_name" should not be in the loop. Why are you talking about validations?

chassis.go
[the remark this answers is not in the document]
hzhou8: No you didn't validate. You just ignored invalid values. You can add a test case to have IP == "" add see how it works.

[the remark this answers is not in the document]
hzhou8: Same for all the other "validation" comments. I'd suggest to add negative test cases ensure the correctness of code.

chassis_encap.go
```diff
@@ -0,0 +1,94 @@
+/**
```
hzhou8: Please follow the convention to use table name as file name. i.e. encap.go

chassis.go
```diff
@@ -133,13 +151,38 @@ func (odbi *ovndb) chassisGetImp(hostname string) ([]*Chassis, error) {
...
+	if !ok {
+		return nil, fmt.Errorf("Chassis with uuid%s not found", uuid)
+	}
 	ch := &Chassis{
-		UUID:       uuid,
-		Name:       odbi.cache[tableChassis][uuid].Fields["name"].(string),
-		Hostname:   odbi.cache[tableChassis][uuid].Fields["hostname"].(string),
-		ExternalID: odbi.cache[tableChassis][uuid].Fields["external_ids"].(libovsdb.OvsMap).GoMap,
-		NbCfg:      odbi.cache[tableChassis][uuid].Fields["nb_cfg"].(int),
+		UUID:                uuid,
```
hzhou8: Still Encaps field is not populated.

chassis.go
```diff
@@ -32,66 +33,79 @@ type Chassis struct {
...
+func (odbi *ovndb) chassisAddImp(name string, hostname string, etype []string, ip string,
+	external_ids map[string]string, transport_zones []string, vtep_lswitches []string) (*OvnCommand, error) {
+	if len(name) == 0 {
+		return nil, fmt.Errorf("chassis name cannot be empty")
 	}
-	row := make(OVNRow)
-
-	if len(name) > 0 {
-		row["chassis_name"] = name
+	if len(etype) == 0 {
```
hzhou8: What if it is nil? Same for all other validations.

chassis.go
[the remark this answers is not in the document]
noah8713: Sorry what is your confusion as we need per raw entries? e.g. when we add 3 encaps for a chassis, we still use same name
_uuid               : 65a59100-40e7-4ead-88b2-15a9a73cd93f
chassis_name        : "ch1"
ip                  : "1.2.3.5"
options             : {csum="true"}
type                : stt

_uuid               : ba8ef8a5-2c9d-4537-8983-b414fc8db520
chassis_name        : "ch1"
ip                  : "1.2.3.5"
options             : {csum="true"}
type                : vxlan

_uuid               : 7f3fc486-1d9e-4b32-89b0-42b705a6275f
chassis_name        : "ch1"
ip                  : "1.2.3.5"
options             : {csum="true"}
type                : geneve

chassis.go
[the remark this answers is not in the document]
noah8713: PTAL https://github.com/eBay/go-ovn/pull/71/files#diff-b79da9a73380e10416689124be77bb72R44

chassis.go
```diff
@@ -32,66 +33,79 @@ type Chassis struct {
...
-	rowChassis["encaps"] = encap_id
-	if len(name) > 0 {
-		rowChassis["name"] = name
+	// consolidate encaps
+	var encaps []interface{}
+	encaps = append(encaps, "set" )
+	encaps = append(encaps, encap_ids)
+	rowChassis["encaps"] = encaps
+    rowChassis["name"] = name
+    rowChassis["hostname"] = hostname
```
hzhou8: It appears to be alignment problem for the above two rows.

chassis_encap_test.go
```diff
@@ -0,0 +1,82 @@
...
+	if err != nil {
+		t.Fatal(err)
+	}
+	if len(chassis) != 1 {
+		t.Fatalf("err getting chassis, total:%v", len(chassis))
+	}
+	chName := chassis[0].Name
+	t.Logf("Chassis found:%+v", chName)
+
+	// Verify if encaps are created
```
hzhou8: Need to verify if encaps are referenced in Chassis's encaps column. You defined it in the Chassis type.

chassis.go
hzhou8: You generated a new UUID, there is no point to check if a new UUID existed. UUID is supposed to be universally unique.

chassis.go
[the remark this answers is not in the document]
hzhou8: You are right. Please ignore this comment.

chassis_encap.go
noah8713: We follow same for other components with what its tied to e.g. logical_router_static_route.go, etc. So kept it. Is it ok?

chassis.go
noah8713: ohk I thought you wanted to add encap read support which I added. Sure i'll add it here too just to populate the uuid!

chassis.go
noah8713: for strings nil cannot be assigned but for array,  it works.
package main

import "fmt"

func main (){
	var etype [] string

	if len(etype) == 0 {
	     fmt.Println("len is 0 for empty array")
	}
	etype = nil
	if len(etype) == 0 {
	     fmt.Println("len is 0 for nil")
	}
}
o/p:
len is 0 for empty array
len is 0 for nil

chassis.go
noah8713: It's not showing in the editor. I double checked spaces

chassis_encap_test.go
noah8713: ok I'll add that UT too

chassis.go
```diff
@@ -134,12 +155,64 @@ func (odbi *ovndb) chassisGetImp(hostname string) ([]*Chassis, error) {
...
+		case libovsdb.OvsSet:
+			ch.VtepLogicalSwitches = odbi.ConvertGoSetToStringArray(vtep.(libovsdb.OvsSet))
+		}
+	}
+	var encaps []string
+	if enc, ok := cacheChassis.Fields["encaps"]; ok {
+		switch enc.(type) {
+		case libovsdb.UUID:
+			if enuid, ok := enc.(libovsdb.UUID); ok {
+				cenc, err := odbi.rowToEncap(enuid.GoUUID)
```
hzhou8: what's the purpose of this rowToEncap()? You convert from UUID to Encap, but later, finally, you just append the UUID to encaps, then why do this conversion at all for the UUID of Encap?

noah8713: You are right. No need of rowToEncap. Taken care in revised commit

PR #72, opened 2020-02-20
Refactor client, api and UTs:

noah8713:
1. Each UT should use its own client
2. Handle client interface to default to nb db if no db
   is passed to ensure old client interface is backward
   compatible.
3. Fix CI to run all UTs accordingly.

### DIFF
--- a/client.go
+++ b/client.go
@@ -17,11 +17,10 @@
 package goovn
 
 import (
+	"fmt"
 	"github.com/ebay/libovsdb"
 	"sync"
 )
-
-var db string
 
 // Client ovnnb/sb client
 // Note: We can create different clients for ovn nb and sb each in future.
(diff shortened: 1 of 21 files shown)


noah8713:
@hzhou8 : PTAL at this too!

client.go
```diff
@@ -180,13 +179,24 @@ type ovndb struct {
...
 
-func NewClient(cfg *Config, db string) (Client, error) {
+func NewClient(cfg *Config) (Client, error) {
+	db := cfg.db
+	// db string should strictly be OVN_Northbound or OVN_Southbound
+	if db == "" {
+		// default to OVN_Northbound
+		db = dbNB
+	} else if !(db == dbNB || db == dbSB) {
+		return nil, fmt.Errorf("db %s is not supported option", db)
```
hzhou8: better to make it clear in the error message: the valid "db" names are: "OVN_Northbound", "OVN_Southbound"

noah8713: done

PR #73, opened 2020-02-25
Make db as exportable

noah8713:
Since new clients can now specify where they want to connect to,
either nb db or sb db, make db identifier as exportable so that
when used in other packages, it can be available to use as is.


noah8713:
Thanks guys for quick review. Merging it as got 2 lgtms.

PR #74, opened 2020-02-27
Make DB constants as exportable

noah8713:
Since clients specify either nb or sb db, client can re-use the constants
from go-ovn directly.
Hence, renaming dbNB as dbSB to make those identifiers explortable.

### DIFF
--- a/client.go
+++ b/client.go
@@ -191,9 +191,9 @@ func NewClient(cfg *Config) (Client, error) {
 	// db string should strictly be OVN_Northbound or OVN_Southbound
 	if db == "" {
 		// default to OVN_Northbound
-		db = dbNB
-	} else if !(db == dbNB || db == dbSB) {
-		return nil, fmt.Errorf("Valid db names are: %s and %s", dbNB, dbSB)
+		db = DBNB
+	} else if !(db == DBNB || db == DBSB) {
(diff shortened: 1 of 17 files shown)


common.go
```diff
@@ -25,8 +25,8 @@ const (
 )
 
 const (
-	dbNB string = "OVN_Northbound"
-	dbSB string = "OVN_Southbound"
+	DbNB string = "OVN_Northbound"
+	DbSB string = "OVN_Southbound"
```
vtolstov: in go usually DB is uppercase, so DBNB/DBSB

client.go
```diff
@@ -191,9 +191,9 @@ func NewClient(cfg *Config) (Client, error) {
 	// db string should strictly be OVN_Northbound or OVN_Southbound
 	if db == "" {
 		// default to OVN_Northbound
-		db = dbNB
```
vtolstov: i'm prefer to rewrite this to switch, like
```
db := cfg.Db
case  db {
    DBNB,DBSB:
        break
    "":
       db = DBNB
    default:
      error
}
```

noah8713: @vtolstov will handle refactoring in separate PR as this commit is just to rename the vars

common.go
noah8713: ok done!

noah8713:
Thanks guys for review!

PR #75, opened 2020-02-28
Move db validation to switch case

noah8713:


client.go
```diff
@@ -189,6 +189,15 @@ type ovndb struct {
...
 	// db string should strictly be OVN_Northbound or OVN_Southbound
+	switch db {
+	case DBNB, DBSB:
+		break
+	case "":
+		db = DBNB
+	default:
+		return nil, fmt.Errorf("Valid db names are: %s and %s", DBNB, DBSB)
+	}
+
```
hzhou8: The change doesn't seem necessary. Meanwhile, L201-206, the old validation, were not removed.

vtolstov: why? its more readable now

noah8713: @hzhou8 :Yes removed l201-206. My bad in testing!

hzhou8: > why? its more readable now

Well, it doesn't look more (or less) readable to me, but it is good if at least it is more readable to you :)

noah8713:
Thanks for the review guys!

PR #77, opened 2020-03-20
Fix lr static route client error

noah8713:
When listing static routes for a logical router, ensure router exist in ovn.
Current code returns empty static routes list without any error.
Hence, this commit fixes it.

### DIFF
--- a/logical_router_static_route.go
+++ b/logical_router_static_route.go
@@ -18,7 +18,6 @@ package goovn
 
 import (
 	"fmt"
-
 	"github.com/ebay/libovsdb"
 )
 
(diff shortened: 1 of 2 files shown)


logical_router_static_route.go
```diff
@@ -157,7 +156,7 @@ func (odbi *ovndb) lrsrListImp(lr string) ([]*LogicalRouterStaticRoute, error) {
 	if !ok {
 		return nil, ErrorNotFound
 	}
-
+	var lrFound bool
```
vtolstov: why we need additional flag?

logical_router_static_route.go
```diff
@@ -185,8 +184,13 @@ func (odbi *ovndb) lrsrListImp(lr string) ([]*LogicalRouterStaticRoute, error) {
...
 				}
 			}
+			lrFound = true
 			break
 		}
 	}
+
+	if !lrFound {
+		return nil, ErrorNotFound
+	}
```
vtolstov: why not simply check 
if len(listLRSR) == 0 {
  return nil, ErrorNotFound
}

logical_router_static_route.go
noah8713: because we iterate through all the lrs and update this to found = true if lr exists in ovn.

logical_router_static_route.go
noah8713: Because it is ok to not have any lsrs for a given router.

hzhou8: agree. this is better ;)

PR #79, opened 2020-03-25
Fix missing parameters of execution function of OvnCommand

fyuan1316:
# add the missing parameters, so we can execute ovnCommand directly.

Before:
---
ocmd, _ := ovndbapi.LSAdd("ls1")
ovndbapi.Execute(ocmd)

After:
---
ocmd, _ := ovndbapi.LSAdd("ls1")
ocmd.Execute()

### DIFF
--- a/client.go
+++ b/client.go
@@ -176,6 +176,8 @@ type Client interface {
 	Close() error
 }
 
+var _ Client = &ovndb{}
+
 type ovndb struct {
 	client       *libovsdb.OvsdbClient
 	cache        map[string]map[string]libovsdb.Row
--- a/apis.go
+++ b/apis.go
@@ -29,7 +29,7 @@ type OvnCommand struct {
 
 // Execute sends command to ovnnb
 func (ocmd *OvnCommand) Execute() error {
-	return ocmd.Exe.Execute()
+	return ocmd.Exe.Execute(ocmd)
 }
 
 // Execution executes multiple ovnnb commands
--- a/examples/main.go
+++ b/examples/main.go
@@ -46,10 +46,14 @@ func init() {
 
 func main() {
 
+	// execute a ovnCommand by either of the following two methods :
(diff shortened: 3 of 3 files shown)


fyuan1316:
@hzhou8  @noah8713 
It looks like the parameters are forgotten, if this is the original design, please correct me  ：）

fyuan1316:
add example and some description ：）

noah8713:
@fyuan1316 : Did CI pass for this change?

client.go
```diff
@@ -176,6 +176,8 @@ type Client interface {
 	Close() error
 }
 
+var _ Client = &ovndb{}
```
hzhou8: Why this change?

fyuan1316: Just to ensure that the `ovndb` struct  implements the methods defined by the `Client` interface

noah8713:
Thanks for the explaination!

hzhou8:
This PR should be squashed before merge.

noah8713:
This PR had two commits in same PR. Merge didn't complain for rebase so I merged it instead of squash.

hzhou8:
@noah8713 The second commit is a fix of the previous commit (a revision based on review comment), which is why they should be squashed. In another words, they belong to same change, and the principles is one commit for one change.

noah8713:
@hzhou8 : I just checked the commits. Since commit https://github.com/eBay/go-ovn/pull/79/commits/40b307c93fbdf702fba1698989bf46b3d814cb0d is the one on top of https://github.com/eBay/go-ovn/pull/79/commits/0f9821be95ba27b9273b028633a249d23cbe6b25 , review comments code got addressed fine. So ideally what you are proposing is user should be sending one PR for each commit instead of two commits in same PR. Do you want me to revert the merge? We can ask user to send one PR with both two commits squashed into one.

PR #82, opened 2020-04-02
Support for LSPGet API in the client interface as well as configuration of LSP fields

abhat:
This pull request adds support for 

1) the LSPGet API in the client's interface. The implementation is already present, but the API was not being exposed.

2) dynamic address configuration for the logical switch port.

3) support for configuration of external ids on the logical switch port.

Signed-off-by: Aniket Bhat <anbhat@redhat.com>

### DIFF
--- a/client.go
+++ b/client.go
@@ -18,8 +18,9 @@ package goovn
 
 import (
 	"fmt"
-	"github.com/ebay/libovsdb"
 	"sync"
+
+	"github.com/ebay/libovsdb"
 )
 
 // Client ovnnb/sb client
@@ -40,6 +41,8 @@ type Client interface {
 	// Link logical switch to router
 	LinkSwitchToRouter(lsw, lsp, lr, lrp, lrpMac string, networks []string, externalIds map[string]string) (*OvnCommand, error)
 
+	// Get logical switch port by name
+	LSPGet(lsp string) (*LogicalSwitchPort, error)
 	// Add logical port PORT on SWITCH
 	LSPAdd(ls string, lsp string) (*OvnCommand, error)
 	// Delete PORT from its attached switch
@@ -266,6 +269,10 @@ func (c *ovndb) LSExtIdsAdd(ls string, external_ids map[string]string) (*OvnComm
 
 func (c *ovndb) LSExtIdsDel(ls string, external_ids map[string]string) (*OvnCommand, error) {
 	return c.lsExtIdsDelImp(ls, external_ids)
+}
+
+func (c *ovndb) LSPGet(lsp string) (*LogicalSwitchPort, error) {
+	return c.lspGetImp(lsp)
 }
 
 func (c *ovndb) LSPAdd(ls string, lsp string) (*OvnCommand, error) {
--- a/client.go
+++ b/client.go
@@ -128,7 +128,10 @@ type Client interface {
 	LSPGetDHCPv6Options(lsp string) (*DHCPOptions, error)
 	// Set options in LSP
 	LSPSetOptions(lsp string, options map[string]string) (*OvnCommand, error)
-
+	// Set dynamic addresses in LSP
+	LSPSetDynamicAddresses(lsp string, address string) (*OvnCommand, error)
+	// Get dynamic addresses from LSP
+	LSPGetDynamicAddresses(lsp string) (string, error)
 	// Add dhcp options for cidr and provided external_ids
 	DHCPOptionsAdd(cidr string, options map[string]string, external_ids map[string]string) (*OvnCommand, error)
 	// Set dhcp options and set external_ids for specific uuid
@@ -313,6 +316,14 @@ func (c *ovndb) LSPGetDHCPv6Options(lsp string) (*DHCPOptions, error) {
 
 func (c *ovndb) LSPSetOptions(lsp string, options map[string]string) (*OvnCommand, error) {
 	return c.lspSetOptionsImp(lsp, options)
+}
+
+func (c *ovndb) LSPSetDynamicAddresses(lsp string, address string) (*OvnCommand, error) {
+	return c.lspSetDynamicAddressesImp(lsp, address)
+}
+
+func (c *ovndb) LSPGetDynamicAddresses(lsp string) (string, error) {
+	return c.lspGetDynamicAddressesImp(lsp)
 }
 
 func (c *ovndb) LSLBAdd(ls string, lb string) (*OvnCommand, error) {
--- a/client.go
+++ b/client.go
@@ -132,6 +132,10 @@ type Client interface {
 	LSPSetDynamicAddresses(lsp string, address string) (*OvnCommand, error)
 	// Get dynamic addresses from LSP
 	LSPGetDynamicAddresses(lsp string) (string, error)
+	// Set external_ids for LSP
+	LSPSetExternalIds(lsp string, external_ids map[string]string) (*OvnCommand, error)
+	// Get external_ids from LSP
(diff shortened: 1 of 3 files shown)


abhat:
Also, @noah8713 is there a slack/irc channel where folks working on the library communicate?

noah8713:
> Also, @noah8713 is there a slack/irc channel where folks working on the library communicate?

Not yet. @hzhou8  can comment if there are plans to start one.

logical_switch_port.go
```diff
@@ -206,6 +207,28 @@ func (odbi *ovndb) lspSetOptionsImp(lsp string, options map[string]string) (*Ovn
 	return &OvnCommand{operations, odbi, make([][]map[string]interface{}, len(operations))}, nil
 }
 
+func (odbi *ovndb) lspSetDynamicAddressesImp(lsp string, address string) (*OvnCommand, error) {
```
noah8713, line 210: Addresses is string [] not just string.
Also add a nil check for at-least lsp as its a mandatory param for this operation. 
Double confirm if addresses is also mandatory and hence no harm in adding nil check for that too.

abhat: This is for dynamic_addresses column in the `Logical_Switch_Port` table. That is not a `[]string`, but just a `string`.

abhat: By nil check, you mean checking for empty string? For the lsp name, if someone passes an empty string, I would expect the row fetch to fail. Also, we have a use-case for the dynamic_address field to be set to empty string. So that check isn't needed. Passing an empty string is equivalent of clearing the field.

noah8713: I compared the nbctl commands and it takes just string so we are good.

yes no need to even look in cache for empty string similar to len(lsp)==0 return. We do it for all required entities which lsp in this case.

logical_switch_port_test.go
```diff
@@ -0,0 +1,123 @@
...
+	err = ovndbapi.Execute(cmd)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	lsp2, err := ovndbapi.LSPGet(PORT_TEST_LSP2)
+	if err != nil {
+		t.Fatal(err)
+	}
+	assert.Equal(t, lsp2.DynamicAddresses, PORT_TEST_LSP2DYNADDR2)
```
noah8713, line 121: Finally , also clean up the ls PORT_TEST_LS1 using LSDel

abhat:
@noah8713 thanks for the reviews. Also, added support for configuring the `external_ids` for the LSP. PTAL.

client.go
```diff
@@ -128,7 +128,10 @@ type Client interface {
 	LSPGetDHCPv6Options(lsp string) (*DHCPOptions, error)
 	// Set options in LSP
 	LSPSetOptions(lsp string, options map[string]string) (*OvnCommand, error)
-
+	// Set dynamic addresses in LSP
+	LSPSetDynamicAddresses(lsp string, address string) (*OvnCommand, error)
+	// Get dynamic addresses from LSP
+	LSPGetDynamicAddresses(lsp string) (string, error)
```
hzhou8, line 134: I think the "LSPGetDynamicAddresses" is not necessary. If you get the LSP struct with LSPGet(), you can access any member of it directly. Any benefit of adding this interface separately?

client.go
```diff
@@ -132,6 +132,10 @@ type Client interface {
 	LSPSetDynamicAddresses(lsp string, address string) (*OvnCommand, error)
 	// Get dynamic addresses from LSP
 	LSPGetDynamicAddresses(lsp string) (string, error)
+	// Set external_ids for LSP
+	LSPSetExternalIds(lsp string, external_ids map[string]string) (*OvnCommand, error)
+	// Get external_ids from LSP
+	LSPGetExternalIds(lsp string) (map[string]string, error)
```
hzhou8, line 138: Same here, no need to add the Get interface for external_ids

client.go
abhat: `GetDynamicAddresses` is just a convenience function.

client.go
abhat: `GetExternalIds` is essential because external_ids are saved as `string[interface{}]interface{}`. The implementation returns a map[string]string. It's implementing some type conversion logic.

abhat:
> Thanks for the improvement. In addition to the inlined comments, it seems the 3rd commit is a fix (and tests) for the previous ones. Please consider squashing them in previous commits before merge.

Will squash commits if everything else looks good.

client.go
abhat: Note that `GetDynamicAddresses` also folds `libovsdb.OvsSet` into a single string as expected by the nbdb spec.

hzhou8: ok

client.go
hzhou8: ok

hzhou8:
LGTM except squashing 3rd commit to 1st and 2nd.

abhat:
> LGTM except squashing 3rd commit to 1st and 2nd.

Done. Please merge as appropriate. Also, unrelated, do you/your team hang out on any ovn/ovs IRC/Slack channels to get more "real time" feedback and help with PRs?

hzhou8:
LGTM @noah8713 @vtolstov could any of you confirm as well?

hzhou8:
> do you/your team hang out on any ovn/ovs IRC/Slack channels to get more "real time" feedback and help with PRs?

Yes there is a slack channel, but only me and @vtolstov used it for quite a short while. Feel free to join: https://join.slack.com/t/go-ovn/shared_invite/zt-ddsmtvza-ePqwvAXXzip3HBO62Z1yzA

logical_switch_port.go
```diff
@@ -206,6 +207,28 @@ func (odbi *ovndb) lspSetOptionsImp(lsp string, options map[string]string) (*Ovn
...
+	row := make(OVNRow)
```
noah8713, line 212: Thanks for all the corrections. I think checking lsp is not null is still missed e.g. you can use chassis.go for reference https://github.com/eBay/go-ovn/blob/master/chassis.go#L38 where mandatory field name is always being checked before mutating ovn dbs. Kindly address corresponding UT for null LSP too?

logical_switch_port.go
```diff
@@ -229,6 +230,39 @@ func (odbi *ovndb) lspGetDynamicAddressesImp(lsp string) (string, error) {
 	return lp.DynamicAddresses, nil
 }
 
+func (odbi *ovndb) lspSetExternalIdsImp(lsp string, external_ids map[string]string) (*OvnCommand, error) {
+	mutatemap, _ := libovsdb.NewOvsMap(external_ids)
```
noah8713, line 234: Also add null check for lsp which is mandatory here.

logical_switch_port.go
abhat: I have addressed the two places where I added new methods. But in the rest of the file, all the other existing functions like `lspAdd`, `lspDel`, `lspSetAddress` don't check for an empty string for the lsp. It makes sense to fail faster though than invoking the db transaction and failing.

abhat: I can create a separate PR to address checking for empty strings for lsp name in other implementation methods for the LSP.

abhat:
Squashed commits and addressed the last review comments from @noah8713. PTAL and merge as appropriate. As mentioned above, I will add empty string checks for LSP name in other API implementation methods in a separate PR.

vtolstov:
lgtm, i think that other improvements can be in separate pr

PR #83, opened 2020-04-11
Remove spurious dynamic address type assertion

abhat:
There was a left-over type assertion after changes were made to do a switch for the contained type of the interface. This commit fixes that.

### DIFF
--- a/logical_switch_port.go
+++ b/logical_switch_port.go
@@ -331,7 +331,6 @@ func (odbi *ovndb) rowToLogicalPort(uuid string) *LogicalSwitchPort {
 		default:
 			//	glog.V(OVNLOGLEVEL).Info("Unsupport type found in lport dynamic address.")
 		}
-		lp.DynamicAddresses = dynamicAddresses.(string)
 	}
 
 	return lp
--- a/logical_switch_port.go
+++ b/logical_switch_port.go
@@ -271,7 +271,7 @@ func (odbi *ovndb) lspGetExternalIdsImp(lsp string) (map[string]string, error) {
 	return extIds, nil
 }
 
-func (odbi *ovndb) rowToLogicalPort(uuid string) *LogicalSwitchPort {
+func (odbi *ovndb) rowToLogicalPort(uuid string) (*LogicalSwitchPort, error) {
 	lp := &LogicalSwitchPort{
 		UUID:       uuid,
 		Name:       odbi.cache[tableLogicalSwitchPort][uuid].Fields["name"].(string),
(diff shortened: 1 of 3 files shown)


logical_switch_port.go
```diff
@@ -331,7 +331,6 @@ func (odbi *ovndb) rowToLogicalPort(uuid string) *LogicalSwitchPort {
 		default:
 			//	glog.V(OVNLOGLEVEL).Info("Unsupport type found in lport dynamic address.")
 		}
-		lp.DynamicAddresses = dynamicAddresses.(string)
```
noah8713, line 334: LGTM for this part.

logical_switch_port.go
```diff
@@ -331,7 +331,6 @@ func (odbi *ovndb) rowToLogicalPort(uuid string) *LogicalSwitchPort {
 		default:
 			//	glog.V(OVNLOGLEVEL).Info("Unsupport type found in lport dynamic address.")
```
noah8713, line 332: Can you also get rid of   unwanted comment //	glog.V(OVNLOGLEVEL) and return error for unsupported type instead
return nil, fmt.Errorf("Unsupport type found in lport dynamic address.")
. Seems we missed this in previous review.

abhat:
@hzhou8, @noah8713 can one of you merge the PR?

logical_switch_port.go
```diff
@@ -303,7 +303,7 @@ func (odbi *ovndb) rowToLogicalPort(uuid string) *LogicalSwitchPort {
 		case libovsdb.OvsSet:
 			lp.Addresses = odbi.ConvertGoSetToStringArray(addr.(libovsdb.OvsSet))
 		default:
-			//	glog.V(OVNLOGLEVEL).Info("Unsupport type found in lport address.")
```
hzhou8, line 306: Originally, the function silently ignores the error for the column data it cannot handle. Now this commit changes the behavior to fail the whole conversion.

Maybe it is better to print an error about the conversion failure but still handle the rest of the data as best as it can. Not sure why the log was commented out originally. @softwarejl do you remember this?

The only scenario this could happen is when the server side schema changed but the go-ovn lib is not up-to-date. I think maybe printing an error log is better than failing the whole function.

This commit is not directly related to the original PR. The first commit of the PR LGTM.

softwarejl: when wrote it, it's weird for me, why the data can be reflect to different struct.  so leave a default case line there, just worried about if some type is not covered in some version.

noah8713: @hzhou8 : To add on;  comment //glog.V part was introduced by current master commit 8710ab5d55e2d944c59b81b2d30b9618b9c18b8b  after adding dynamic addressing support by @abhat. Hence he submitted two commits as per my understanding to fix the same. Correct me if I am wrong @abhat.

abhat: @noah8713 the commented line was introduced by my previous PR, but it is following the pattern originally present with the other LSP fields in `rowToLogicalPort`. 

Having said that, I am personally perfectly OK to have appropriate error handling as you suggested. Accordingly, all calls to `rowToLogicalPort` have been modified. It's better to return a nil row instead of returning partial data from the row. In all valid cases of schema-adhering  fields, the default switch case shouldn't be executed.

hzhou8: @abhat I am ok with your change (just wanted to make sure this was thought through). Thanks a lot!

PR #84, opened 2020-04-22
Add api support for NB/SB Global tables and fix the LSP APIs for options and external ids.

abhat:
This commit adds support for creating and deleting NB/SB rows via the go-api as also the ability to set options on the rows. In reality the add/delete rows is only used for unit-testing since ovn-northd is the one that creates the rows during start up in the NB_Global table.

There was also a bug in the way the LSP SetOptions and SetExternalIds api worked. Updating entries in these maps would fail since instead of calling an `opUpdate` on the entire row, these APIs were trying to mutate options/external-ids using individual key-values using `opInsert`. That would have failed unless there is a delete first followed by an insert.

### DIFF
--- a/logical_switch_port.go
+++ b/logical_switch_port.go
@@ -193,19 +193,50 @@ func (odbi *ovndb) lspGetDHCPv6OptionsImp(lsp string) (*DHCPOptions, error) {
 }
 
 func (odbi *ovndb) lspSetOptionsImp(lsp string, options map[string]string) (*OvnCommand, error) {
-	mutatemap, _ := libovsdb.NewOvsMap(options)
-	mutation := libovsdb.NewMutation("options", opInsert, mutatemap)
+	if options == nil {
+		return nil, ErrorOption
+	}
+
+	if len(lsp) == 0 {
+		return nil, fmt.Errorf("LSP name cannot be empty while setting options")
+	}
+
+	optionsMap, err := libovsdb.NewOvsMap(options)
+	if err != nil {
+		return nil, err
+	}
+
+	row := make(OVNRow)
+	row["options"] = optionsMap
+
 	condition := libovsdb.NewCondition("name", "==", lsp)
 
 	// simple mutate operation
-	mutateOp := libovsdb.Operation{
-		Op:        opMutate,
-		Table:     tableLogicalSwitchPort,
-		Mutations: []interface{}{mutation},
-		Where:     []interface{}{condition},
+	updateOp := libovsdb.Operation{
+		Op:    opUpdate,
+		Table: tableLogicalSwitchPort,
+		Row:   row,
+		Where: []interface{}{condition},
 	}
-	operations := []libovsdb.Operation{mutateOp}
+	operations := []libovsdb.Operation{updateOp}
 	return &OvnCommand{operations, odbi, make([][]map[string]interface{}, len(operations))}, nil
+}
+
+func (odbi *ovndb) lspGetOptionsImp(lsp string) (map[string]string, error) {
+	lp, err := odbi.lspGetImp(lsp)
+	if err != nil {
+		return nil, err
+	}
+	options := make(map[string]string)
+	for k, v := range lp.Options {
+		key, keyOk := k.(string)
+		value, valueOk := v.(string)
+		if !keyOk || !valueOk {
+			continue
+		}
+		options[key] = value
+	}
+	return options, nil
 }
 
 func (odbi *ovndb) lspSetDynamicAddressesImp(lsp string, address string) (*OvnCommand, error) {
@@ -235,22 +266,32 @@ func (odbi *ovndb) lspGetDynamicAddressesImp(lsp string) (string, error) {
 }
 
 func (odbi *ovndb) lspSetExternalIdsImp(lsp string, external_ids map[string]string) (*OvnCommand, error) {
-	if len(lsp) == 0 {
-		return nil, fmt.Errorf("LSP name cannot be empty while setting external ids")
+	if external_ids == nil {
+		return nil, ErrorOption
 	}
 
-	mutatemap, _ := libovsdb.NewOvsMap(external_ids)
-	mutation := libovsdb.NewMutation("external_ids", opInsert, mutatemap)
+	if len(lsp) == 0 {
+		return nil, fmt.Errorf("LSP name cannot be empty while setting external_ids")
+	}
+
+	externalIdsMap, err := libovsdb.NewOvsMap(external_ids)
+	if err != nil {
+		return nil, err
+	}
+
+	row := make(OVNRow)
+	row["external_ids"] = externalIdsMap
+
 	condition := libovsdb.NewCondition("name", "==", lsp)
 
 	// simple mutate operation
-	mutateOp := libovsdb.Operation{
-		Op:        opMutate,
-		Table:     tableLogicalSwitchPort,
-		Mutations: []interface{}{mutation},
-		Where:     []interface{}{condition},
+	updateOp := libovsdb.Operation{
+		Op:    opUpdate,
+		Table: tableLogicalSwitchPort,
+		Row:   row,
+		Where: []interface{}{condition},
 	}
-	operations := []libovsdb.Operation{mutateOp}
+	operations := []libovsdb.Operation{updateOp}
 	return &OvnCommand{operations, odbi, make([][]map[string]interface{}, len(operations))}, nil
 }
 
--- a/logical_switch_port_test.go
+++ b/logical_switch_port_test.go
@@ -31,6 +31,14 @@ const (
 	PORT_TEST_LSP2DYNADDR2   = ""
 	PORT_TEST_EXT_ID_MAC_KEY = "mac_addr"
 	PORT_TEST_EXT_ID_MAC     = "00:01:02:03:04:05"
+	PORT_TEST_EXT_ID_MAC_2   = "01:02:03:05:05:06"
+	PORT_TEST_EXT_ID_IP_KEY  = "ip_addr"
+	PORT_TEST_EXT_ID_IP      = "169.254.1.1"
+	PORT_TEST_OPT_1_KEY      = "foo1"
+	PORT_TEST_OPT_1_VAL      = "bar1"
+	PORT_TEST_OPT_1_VAL_2    = "baz1"
+	PORT_TEST_OPT_2_KEY      = "foo2"
+	PORT_TEST_OPT_2_VAL      = "bar2"
 )
 
 func TestLogicalSwitchPortAPI(t *testing.T) {
@@ -123,10 +131,13 @@ func TestLogicalSwitchPortAPI(t *testing.T) {
 	}
 	assert.Equal(t, lsp2.DynamicAddresses, PORT_TEST_LSP2DYNADDR2)
 
+	// test external id APIs
 	extIds, err := ovndbapi.LSPGetExternalIds(PORT_TEST_LSP1)
 	assert.Equal(t, extIds != nil, true)
 
+	t.Logf("Setting external ids with two keys")
 	extIds[PORT_TEST_EXT_ID_MAC_KEY] = PORT_TEST_EXT_ID_MAC
+	extIds[PORT_TEST_EXT_ID_IP_KEY] = PORT_TEST_EXT_ID_IP
 
 	cmd, err = ovndbapi.LSPSetExternalIds(PORT_TEST_LSP1, extIds)
 	if err != nil {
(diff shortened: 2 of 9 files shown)


nb_sb_global.go
```diff
@@ -0,0 +1,186 @@
...
+
+package goovn
+
+import (
+	"fmt"
+
+	"github.com/ebay/libovsdb"
+)
+
+type NBGlobalTableRow struct {
```
abhat, line 25: These are types defined mostly for future use. They can be used to get back the entire row in the future.

nb_sb_global.go
```diff
@@ -0,0 +1,186 @@
...
+func (odbi *ovndb) globalSetOptionsImp(options map[string]string, table string) (*OvnCommand, error) {
+	if options == nil || table == "" {
+		return nil, fmt.Errorf("Invalid arguments passed to set options: table: %s, options:  %v", table, options)
+	}
+	mutatemap, err := libovsdb.NewOvsMap(options)
+	if err != nil {
+		return nil, err
+	}
+
+	uuid, err := func() (string, error) {
```
abhat, line 128: Assumes that the NB_Global and SB_Global tables can have at most one row as per the spec.

nb_sb_global.go
noah8713: @hzhou8 : Can confirm as in the past we had a discussion on this to have each go file for each entity wherein nb_global and sb_global has to be different go file to be consistent with current code.

nb_sb_global.go
noah8713: Can you add a comment about how nbctl/sbctl does as an example for ease to understand.

nb_sb_global_test.go
```diff
@@ -0,0 +1,124 @@
+/**
+ * Copyright (c) 2019 eBay Inc.
```
noah8713, line 2: 2020

logical_switch_port.go
```diff
@@ -193,16 +193,30 @@ func (odbi *ovndb) lspGetDHCPv6OptionsImp(lsp string) (*DHCPOptions, error) {
 }
 
 func (odbi *ovndb) lspSetOptionsImp(lsp string, options map[string]string) (*OvnCommand, error) {
-	mutatemap, _ := libovsdb.NewOvsMap(options)
-	mutation := libovsdb.NewMutation("options", opInsert, mutatemap)
+	if options == nil {
+		return nil, ErrorOption
+	}
+
+	if len(lsp) == 0 {
```
noah8713, line 200: Why include lsp validations when adding nb/sb global db support? Can you move this validation in separate commit?

nb_sb_global.go
```diff
@@ -0,0 +1,186 @@
+/**
+ * Copyright (c) 2017 eBay Inc.
```
noah8713, line 2: 2020

noah8713:
Many UTs are failing on this current patch: https://travis-ci.org/github/eBay/go-ovn/jobs/677990243 . Can you please fix them and amend the patch with correction?

nb_sb_global.go
abhat: @hzhou8 any preferences on splitting the implementation into two files?

logical_switch_port.go
abhat: Will do. Ack.

abhat: Btw, these are not just validations. The implementation of the `SetOptions` and `SetExternalIds` was not correct. The mutate operation would fail if the map argument contained options/external ids already present in the db.

logical_switch_port.go
```diff
@@ -235,20 +249,30 @@ func (odbi *ovndb) lspGetDynamicAddressesImp(lsp string) (string, error) {
...
+
 	condition := libovsdb.NewCondition("name", "==", lsp)
 
 	// simple mutate operation
 	mutateOp := libovsdb.Operation{
-		Op:        opMutate,
-		Table:     tableLogicalSwitchPort,
-		Mutations: []interface{}{mutation},
-		Where:     []interface{}{condition},
+		Op:    opUpdate,
```
hzhou8, line 272: This way, it changes the API behavior. Originally, the API only adds key-values on top of existing ones. Now it would completely replace the old ones. E.g. if there are k1:v1, k2:v2 existed, but the API caller just set k3:v3, then in the end k1 and k2 will both disappear. I am not sure if this is the desired behavior of all existing interface use cases. If using update, it is better to:

1) read old key-values from cache

2) use "wait" operation to make sure it is not changed by anyone else in parallel

3) add the new k-v to existing map and update

You can check ovn-nbctl -v set ... external-ids:xxx=yyy to see the transaction details.

Alternatively, we can use mutation, but delete the old entries if existed.

abhat: Yes, @hzhou8. But the original implementation was not correct. In particular, it will barf if the argument map contained existing keys. I agree this is a behavior change. But the Set API should clobber the existing map entries IMO. We can provide a separate AddOptions/AddExternalIds API to do the diff (i.e. the correct implementation of existing API logic).

abhat: I don't know how many people were using the existing Set API to semantically mean that this is an additive change.

hzhou8: @abhat make sense. The function name "set" suggests a full update. Not sure about impact for existing users. Maybe @vtolstov @softwarejl @noah8713 can comment. If there is no impact for this behavior change for your current use cases, I'd support change it to completely replace the whole column with this API. It is similar for lsp_set_options API.

abhat:
> P.S. the test failure may be due to the fact that OVN DBs are initiated with the row in Global table, so you can't add one if it existed. Maybe the API of adding the row in Global table is not quite useful.

Yes, I didn't realize that the OVN DBs are initiate with rows in Global table. That violates the schema constraint of having only one row. Do you know who is adding that row in the Global table? I know a row is added by northd if it doesn't exist.

logical_switch_port.go
noah8713: Agree with the update part. However, better confirm the current UT works with mutateOp and then add a new patch to cover add/set operations both. However, better add addextid support too so that for both set/add,  fetching from cache odbi.cache[][uuid].Fields["external_ids"].(libovsdb.OvsMap).GoMap is needed  and then update the k:v that needs to be set/add/ . I did confirm with nbctl command it seems idl takes all keys even if one key is being set:
existing external_ids: {aa=yyy, abc="456", xxx="122"}
Run to update xxx=123:. ovn-nbctl -v set logical_switch_port sw0-port1 external-ids:xxx=123
verbose-data: "external_ids":["map",[["aa","yyy"],["abc","456"],["xxx","123"]]]}}
For go client, we might need to check if we need to fetch from cache and explicitely update the key.

abhat: I can implement the`AddExternalIds`/`AddOptions" API in a separate PR since that doesn't have any bearing with the changes here.

To summarize: 

SetOptions/SetExternalIds will do a full-update of the columns in question.

AddOptions/AddExternalIds will do a logical union of keys including mutating the intersecting keys' values.

Assuming we are in agreement here, I have updated the PR and fixed the failing UTs. I have also addressed @noah8713's previous comments. PTAL. I would like for this PR to get merged including the fix to not recursively call clearConnection in the libovsdb repo: https://github.com/eBay/libovsdb/pull/18.

logical_switch_port_test.go
```diff
@@ -149,11 +149,11 @@ func TestLogicalSwitchPortAPI(t *testing.T) {
 
 	//validate that setting fields on an empty LSP string gives a nil cmd and an error
 	cmd, err = ovndbapi.LSPSetDynamicAddresses(PORT_TEST_LSP3, PORT_TEST_LSP1DYNADDR1)
-	assert.Equal(t, cmd, nil)
+	assert.Equal(t, cmd == nil, true)
 	assert.Equal(t, err == nil, false)
 
 	cmd, err = ovndbapi.LSPSetExternalIds(PORT_TEST_LSP3, extIds)
-	assert.Equal(t, cmd, nil)
+	assert.Equal(t, cmd == nil, true)
```
noah8713, line 156: Thanks @abhat for addressing all the comments and I think I am on same page for sets.
Considering that, I have one final comment to cover the full row update scenario. Can you also cover the same in the UTs by setting one more foo:bar value in externalids for PORT_TEST_LSP1 and make sure getextid is returning multiple keys and not just PORT_TEST_EXT_ID_MAC_KEY after https://github.com/eBay/go-ovn/blob/master/logical_switch_port_test.go#L148. Same for setOptions too to make sure update is including old+new key and not excluding old key.

abhat: Done.

logical_switch_port_test.go
```diff
@@ -147,13 +158,119 @@ func TestLogicalSwitchPortAPI(t *testing.T) {
...
+	extIdIP, ok := extIdsRet[PORT_TEST_EXT_ID_IP_KEY]
+	assert.Equal(t, ok, true)
+	assert.Equal(t, extIdIP, PORT_TEST_EXT_ID_IP)
+
+	t.Logf("Validated the external ids are set correctly")
+	// update one of the existing keys, remove one key
+	// and make sure it's a complete update.
+	t.Logf("Validating that keys get clobbered by Set API correctly")
+	extIds[PORT_TEST_EXT_ID_MAC_KEY] = PORT_TEST_EXT_ID_MAC_2
+	delete(extIds, PORT_TEST_EXT_ID_IP_KEY)
```
noah8713, line 170: Hi @abhat : Sorry may I couldn't point correctly. What I meant is if you set extids/options for a k:v , existing k:v should not be impacted.
If you want to delete the extIds/Options, do it with delAPI not with just dict delete(). e.g. if there are two extIDs: a:b , c:d, if you set c=f, a:b should be present too. That is why I gave the e.g. of nbctl :
e.g `current extIDs: external_ids        : {aa=yyy, abc="456", jjj="777", xxx="123", zzz="123"}`

`run ovn-nbctl  set logical_switch_port sw0-port1 external-ids:zzz=125`

`final output: external_ids        : {aa=yyy, abc="456", jjj="777", xxx="123", zzz="125"}`
You can also refer to dhcp_options.go where it implements Add,set,del for options for ref. In this PR, we don't even have add/del api support for extid yet and only set. Hence, you keep setting multiple Keys but other keys should be intact.
Also for Setting options/ext ID, don't do get first. You can set it with map. e.g. refer to dhcp_options_test.go for ref. Correct me if I am wrong.

abhat: I think you are missing the semantic of `SetOptions` or `SetExternalIds`. I thought we agreed that a Set API should do a clobbering of existing keys. And an `AddOption`/`DelOption` API should do the addition/deletion of individual keys.

abhat: The GetOptions or GetExternalIds in the unit-test is just to make sure that the default value for the field in the row is not nil but it's of type OvsMap.

abhat: Here is what @hzhou8 had to say on the issue: https://github.com/ebay/go-ovn/pull/84#discussion_r417015003

logical_switch_port.go
```diff
@@ -235,20 +249,30 @@ func (odbi *ovndb) lspGetDynamicAddressesImp(lsp string) (string, error) {
...
 
-	mutatemap, _ := libovsdb.NewOvsMap(external_ids)
-	mutation := libovsdb.NewMutation("external_ids", opInsert, mutatemap)
+	mutatemap, err := libovsdb.NewOvsMap(external_ids)
+	if err != nil {
+		return nil, err
+	}
+
+	row := make(OVNRow)
+	row["external_ids"] = mutatemap
```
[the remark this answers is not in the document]
abhat: Done.

logical_switch_port_test.go
noah8713: Thanks @abhat . Yup, I align with the discussion that @hzhou8 pointed. But after cross checking ovn-nbctl commands, it conflicts with our consent of full update on the column where set on a key just updates value of existing key no mutation to other keys. If we are still ok with current approach,  it's lgtm from my side. @hzhou8 : Feel free to take a final look and merge if no more suggestions.

hzhou8: Yes, my first comment regarding this was from ovn-nbctl alike command perspective. However, I realized that in the API names convention, set usually means full update. It is better to keep that convention aligned at this lib API level. We can add UpdateOptionsAddKey, UpdateOptionsDelKey, UpdateOptionsUpdateKey, etc. later for the operations where full update is not desired, but again since we have a plan for auto-generating APIs, let's don't waste too much time for operations not really needed. For now I think the API of current patch is good enough.

noah8713: Sounds good. Thanks for confirming.

PR #86, opened 2020-06-02
Support automatically reconnect when disconnected from server.

hzhou8:
This patch implements auto-reconnection, and it can be enabled by
setting Config.Reconnect to true when calling NewClient().

Signed-off-by: Han Zhou <hzhou8@ebay.com>

### DIFF
--- a/client.go
+++ b/client.go
@@ -20,7 +20,10 @@ import (
 	"fmt"
 	"sync"
 
+	"crypto/tls"
 	"github.com/ebay/libovsdb"
+	"log"
+	"time"
 )
 
 // Client ovnnb/sb client
@@ -210,6 +213,25 @@ type ovndb struct {
 	signalCB     OVNSignal
 	disconnectCB OVNDisconnectedCallback
 	db           string
+	addr         string
+	tlsConfig    *tls.Config
+	reconn       bool
+}
+
+func connect(c *ovndb) error {
+	ovsdb, err := libovsdb.Connect(c.addr, c.tlsConfig)
+	if err != nil {
+		return err
+	}
+	c.client = ovsdb
+	initial, err := ovsdb.MonitorAll(c.db, "")
+	if err != nil {
+		return err
+	}
+	c.populateCache(*initial)
+	notifier := ovnNotifier{c}
+	ovsdb.Register(notifier)
+	return nil
 }
 
 func NewClient(cfg *Config) (Client, error) {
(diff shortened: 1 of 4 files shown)


vtolstov:
Looks good

client.go
```diff
@@ -224,29 +246,44 @@ func NewClient(cfg *Config) (Client, error) {
...
-	imp.populateCache(*initial)
-	notifier := ovnNotifier{imp}
-	imp.client.Register(notifier)
-
-	return imp, nil
+	return ovndb, err
+}
+
+func (c *ovndb) reconnect() {
+	ticker := time.NewTicker(500 * time.Millisecond)
```
noah8713, line 267: @hzhou8 : Can we move this logic in libovsdb client https://github.com/eBay/libovsdb/blob/master/client.go#L59 with Reconnect as a new function like libovsdb.Reconnect ? The thing is we will end up re-writing this same reconnect client logic for libovsdb too as this problem is common for ovs/ovn dbs.

ovnimp.go
```diff
@@ -210,7 +210,12 @@ func (odbi *ovndb) populateCache(updates libovsdb.TableUpdates) {
 			odbi.float64_to_int(row.New)
 
 			if !reflect.DeepEqual(row.New, empty) {
+				if reflect.DeepEqual(row.New, odbi.cache[table][uuid]) {
```
abhat, line 213: Do you need to check for `row, ok := odbi.cache[table][uuid]` before calling L213?

ovnnotify.go
```diff
@@ -33,8 +33,11 @@ func (notify ovnNotifier) Stolen([]interface{}) {
 }
 func (notify ovnNotifier) Echo([]interface{}) {
 }
+
 func (notify ovnNotifier) Disconnected(client *libovsdb.OvsdbClient) {
```
softwarejl, line 37: dont need verify client == self?

client.go
hzhou8: I did it first in libovsdb, but then I realized that it is not the right place because the IDL cache is maintained in go-ovn only.

ovnimp.go
hzhou8: I tried that DeepEqual handles it well, DeepEqual(<something not nil>, nil) returns false, so I think it doesn't require an extra check here. What do you think?

ovnnotify.go
hzhou8: It should be ok. libovsdb maintains the connections table for each connection and its CBs, and only calls the CB for the handler of the disconnected client.

PR #87, opened 2020-06-02
delete route by additional parameters

jxiaobin:
add new API to delete route by prefix, nexthop and policy.
nexthop and policy are optional.

### DIFF
--- a/client.go
+++ b/client.go
@@ -97,8 +97,8 @@ type Client interface {
 
 	// Add LRSR with given ip_prefix on given lr
 	LRSRAdd(lr string, ip_prefix string, nexthop string, output_port []string, policy []string, external_ids map[string]string) (*OvnCommand, error)
-	// Delete LRSR with given ip_prefix on given lr
-	LRSRDel(lr string, ip_prefix string) (*OvnCommand, error)
+	// Delete LRSR with given ip_prefix, nexthop, policy and outputPort on given lr
+	LRSRDel(lr string, prefix string, nexthop, policy, outputPort *string) (*OvnCommand, error)
 	// Get all LRSRs by lr
 	LRSRList(lr string) ([]*LogicalRouterStaticRoute, error)
 
@@ -396,8 +396,8 @@ func (c *ovndb) LRSRAdd(lr string, ip_prefix string, nexthop string, output_port
 	return c.lrsrAddImp(lr, ip_prefix, nexthop, output_port, policy, external_ids)
 }
 
-func (c *ovndb) LRSRDel(lr string, ip_prefix string) (*OvnCommand, error) {
-	return c.lrsrDelImp(lr, ip_prefix)
+func (c *ovndb) LRSRDel(lr string, prefix string, nexthop, policy, outputPort *string) (*OvnCommand, error) {
+	return c.lrsrDelImp(lr, prefix, nexthop, policy, outputPort)
 }
 
 func (c *ovndb) LRSRList(lr string) ([]*LogicalRouterStaticRoute, error) {
--- a/logical_router_static_route.go
+++ b/logical_router_static_route.go
@@ -18,6 +18,7 @@ package goovn
 
 import (
 	"fmt"
+
 	"github.com/ebay/libovsdb"
 )
 
@@ -36,6 +37,7 @@ func (odbi *ovndb) lrsrAddImp(lr string, ip_prefix string, nexthop string, outpu
 	if err != nil {
 		return nil, err
 	}
+
 	row := make(OVNRow)
 	row["ip_prefix"] = ip_prefix
 	row["nexthop"] = nexthop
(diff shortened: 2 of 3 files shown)


logical_router_static_route.go
```diff
@@ -36,14 +37,19 @@ func (odbi *ovndb) lrsrAddImp(lr string, ip_prefix string, nexthop string, outpu
...
+	if len(output_port) == 1 {
 		row["output_port"] = output_port
+	} else if len(output_port) > 1 {
+		return nil, fmt.Errorf("only 1 output port is allowed")
 	}
-	if len(policy) > 0 {
-		row["policies"] = policy
+	if len(policy) == 1 {
+		row["policy"] = policy
+	} else {
```
hzhou8, line 51: else if len() > 1

logical_router_static_route.go
```diff
@@ -36,14 +37,19 @@ func (odbi *ovndb) lrsrAddImp(lr string, ip_prefix string, nexthop string, outpu
...
-	if len(output_port) > 0 {
+	if len(output_port) == 1 {
 		row["output_port"] = output_port
+	} else if len(output_port) > 1 {
+		return nil, fmt.Errorf("only 1 output port is allowed")
 	}
-	if len(policy) > 0 {
-		row["policies"] = policy
+	if len(policy) == 1 {
+		row["policy"] = policy
```
hzhou8, line 50: thanks for the fix. This part is better to be in a separate PR.

logical_router_static_route.go
```diff
@@ -115,6 +121,50 @@ func (odbi *ovndb) lrsrDelImp(lr string, ip_prefix string) (*OvnCommand, error)
 	return &OvnCommand{operations, odbi, make([][]map[string]interface{}, len(operations))}, nil
 }
 
+func (odbi *ovndb) lrsrDelImpEx(lr string, prefix string, nexthop, policy *string) (*OvnCommand, error) {
```
hzhou8, line 124: better to rename as lrsrDelExImp

logical_router_static_route.go
hzhou8: Maybe it is better to use the same convention as LRSRDelEx(), i.e. make the parameter *string instead of []string

hzhou8:
Please check CI failure as well

client.go
```diff
@@ -99,6 +99,8 @@ type Client interface {
 	LRSRAdd(lr string, ip_prefix string, nexthop string, output_port []string, policy []string, external_ids map[string]string) (*OvnCommand, error)
 	// Delete LRSR with given ip_prefix on given lr
 	LRSRDel(lr string, ip_prefix string) (*OvnCommand, error)
+	// Delete LRSR with given ip_prefix, nexthop and policy on given lr
+	LRSRDelEx(lr string, prefix string, nexthop, policy *string) (*OvnCommand, error)
```
hzhou8, line 103: Sorry, one more comment here. We should add all parameters including output_port here, as optional parameter, i.e. compare if it is not nil. There can be routes with same prefix, nexthop, policy, but different output_ports. This API should allow user to delete just one of them.

noah8713: Why a need for new API instead of extending the existing LRSRDel? Default can always be delete by prefix if nexthop, policy, etc are not pass.

jxiaobin: @noah8713 we want to selectively delete route with the same prefix but different nexthop, policy.

logical_router_static_route.go
jxiaobin: ok

logical_router_static_route.go
jxiaobin: ok

jxiaobin:
@hzhou8  revised as requested, could you plz take a look?

logical_router_static_route.go
```diff
@@ -115,6 +117,53 @@ func (odbi *ovndb) lrsrDelImp(lr string, ip_prefix string) (*OvnCommand, error)
 	return &OvnCommand{operations, odbi, make([][]map[string]interface{}, len(operations))}, nil
 }
 
+func (odbi *ovndb) lrsrDelExImp(lr string, prefix string, nexthop, policy, outputPort *string) (*OvnCommand, error) {
```
hzhou8, line 120: Sorry, one more comment. It is better that the lrsrDelImp() just call this function now.

client.go
hzhou8: @noah8713 Basically, to avoid breaking existing client for compatibility. If there is no other user using this API, it is better to just change the existing API. @jxiaobin maybe @noah8713 's suggestion makes sense because we just discussed in this PR regarding fixing the API LRSRAdd, which would have to break the existing API users.

logical_router_static_route.go
jxiaobin: fixed

PR #88, opened 2020-06-17
Call reconnect in deferred manner instead of inline with disconnect

abhat:
Make sure that the reconnect call is deferred instead of inline with the disconnect call back.

Signed-off-by: Aniket Bhat <anbhat@redhat.com>

### DIFF
--- a/ovnnotify.go
+++ b/ovnnotify.go
@@ -36,7 +36,7 @@ func (notify ovnNotifier) Echo([]interface{}) {
 
 func (notify ovnNotifier) Disconnected(client *libovsdb.OvsdbClient) {
 	if notify.odbi.reconn {
-		notify.odbi.reconnect()
+		defer notify.odbi.reconnect()
 	} else if notify.odbi.disconnectCB != nil {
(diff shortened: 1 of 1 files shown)


hzhou8:
Thanks for the PR. In libovsdb, clearConnection() would call each handler and then remove the ovsdb client object from the map. At the same time, go-ovn reconnect() is trying to create a new ovsdb client. For my understanding these two steps are independent. Whether the CB in go-ovn is trying to create a new client is transparent to libovsdb. Imaging that another thread (other than the CB) is trying to create a new ovsdb client connecting to the same server in parallel, which should never cause any problem if libovsdb is implemented properly. In current libovsdb implementation, it would simply cause the new client creation get blocked on the mutex of connectionsMutex (https://github.com/eBay/libovsdb/blob/master/client.go#L33), until the CB is scheduled and clearConnection() is completed. So, I think this patch is not necessary, but no harmful either. What your opinion?

hzhou8:
As agreed this is not needed. So close it.

PR #90, opened 2020-07-04
delete static route by uuid

jxiaobin:
allow client to delete static route by uuid

### DIFF
--- a/logical_router_static_route_test.go
+++ b/logical_router_static_route_test.go
@@ -73,14 +73,16 @@ func TestLogicalRouterStaticRoute(t *testing.T) {
 		t.Fatalf("Static Route %s via %s not created in %s", IPPREFIX, nextHop2, LR2)
 	}
 	found := false
+	var secondSr *LogicalRouterStaticRoute
 	for _, sr := range lrsr {
 		if sr.Nexthop == nextHop2 && sr.IPPrefix == IPPREFIX {
 			found = true
+			secondSr = sr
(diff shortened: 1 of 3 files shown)


vtolstov:
LGTM for me, @hzhou8 what you think?

logical_router_static_route_test.go
```diff
@@ -73,14 +73,16 @@ func TestLogicalRouterStaticRoute(t *testing.T) {
...
 	for _, sr := range lrsr {
 		if sr.Nexthop == nextHop2 && sr.IPPrefix == IPPREFIX {
 			found = true
+			secondSr = sr
 		}
 	}
 	assert.Equal(t, true, found, "Added second static route to lr2")
 	// delete static route via nextHop2
-	cmd, err = ovndbapi.LRSRDel(LR2, IPPREFIX, &nextHop2, nil, nil)
+	cmd, err = ovndbapi.LRSRDelByUUID(LR2, secondSr.UUID)
```
hzhou8, line 85: Could you keep the test case for LRSRDel and add a new one for LRSRDelByUUID, instead of deleting the old one?

jxiaobin: @hzhou8 Not all tests for LSRSDel is deleted, there are 2 calls to LRSRDel in this test case, the other call is at line 106/108.

hzhou8: ok

PR #91, opened 2020-07-06
get static route policy

fafucoder:


logical_router_static_route.go
```diff
@@ -158,6 +160,8 @@ func (odbi *ovndb) rowToLogicalRouterStaticRoute(uuid string) *LogicalRouterStat
 			lrsr.OutputPort = []string{outputPort.(libovsdb.UUID).GoUUID}
 		case libovsdb.OvsSet:
 			lrsr.OutputPort = odbi.ConvertGoSetToStringArray(outputPort.(libovsdb.OvsSet))
+		case []string:
```
hzhou8, line 163: Thanks for the PR! This is a bug. It would be better if the commit message could make it clear.

Question: why for Policy it is "case string:" but for output_port it is "case []string:"? In fact, I think both policy and output_port should just be string type, so we should remove the switch-case.

In addition, it would be better to fix the test case as well to cover this.

fafucoder: yes,  policy and output_port should just be string type, but I'm a little confused about why your struct should be set to an array
In fact, i only tested the policy, not the output_port，the next time i will cover this test case

hzhou8: The struct definition was a mistake, too. @jxiaobin was working on a fix, which would change the API, but he stopped the progress. Not sure if he will continue working on it soon.

fafucoder: I've changed the struct and lrsrdel params, which doesn't seem to require pointers

hzhou8:
@fafucoder Sorry for the late response. It seems the PR has conflict now with master. Could you please rebase it?

fafucoder:
Thinks for reply @hzhou8, the conflict has now been resolved and it works fine

hzhou8:
> Thinks for reply @hzhou8, the conflict has now been resolved and it works fine

Unfortunately there is another conflict after @girishmg 's commit. Could you resolve again?

In addition, I think it is better to use type *string instead of string for the optional fields. Otherwise, it is ambiguous when it is "", we don't know if it is "" or empty set in DB. With *string it is clear that nil represents empty, and non nil represents the single element of the set. What do you think?

Finally, could you also update the commit message to describe what's changed in this patch?

fafucoder:
> > 想回复@ hzhou8，现在冲突已解决，并且工作正常
> 
> 不幸的是，@girishmg的提交之后还有另一个冲突。您能再解决吗？
> 
> 此外，我认为最好将* string类型而不是string用于可选字段。否则，它在为“”时会模棱两可，我们不知道它是否为“”或在数据库中设置为空。使用* string可以清楚地看到nil代表空，而非nil代表集合中的单个元素。你怎么看？
> 
> 最后，您是否还可以更新提交消息以描述此补丁中的更改？

Thanks, your suggestion is great！

fafucoder:
> > Thinks for reply @hzhou8, the conflict has now been resolved and it works fine
> 
> Unfortunately there is another conflict after @girishmg 's commit. Could you resolve again?
> 
> In addition, I think it is better to use type *string instead of string for the optional fields. Otherwise, it is ambiguous when it is "", we don't know if it is "" or empty set in DB. With *string it is clear that nil represents empty, and non nil represents the single element of the set. What do you think?
> 
> Finally, could you also update the commit message to describe what's changed in this patch?

Thanks, your suggestion is great！

fafucoder:
I found that @girishmg has solved the problem of not being able to get the policy, so I think this pr is unnecessary

PR #93, opened 2020-07-23
Fix addresses format from slice to string

trozet:
OVN NB expects the addresses to be a single string and not a slice.
Using a slice causes:
invalid syntax '10.128.2.3' in logical switch port addresses. No MAC address found
invalid syntax '10.128.2.3' in port security. No MAC address found

The GoSet object is a slice of addresses instead of a single string:
GoSet:[0a:58:0a:80:04:04 10.128.4.4]

Which causes a newline when we do ovn-nbctl lsp-get-addresses:

0a:58:0a:80:04:03
10.128.4.3

Signed-off-by: Tim Rozet <trozet@redhat.com>

### DIFF
--- a/logical_switch_port.go
+++ b/logical_switch_port.go
@@ -115,11 +115,7 @@ func (odbi *ovndb) lspDelImp(lsp string) (*OvnCommand, error) {
 
 func (odbi *ovndb) lspSetAddressImp(lsp string, addr ...string) (*OvnCommand, error) {
 	row := make(OVNRow)
-	addresses, err := libovsdb.NewOvsSet(addr)
-	if err != nil {
-		return nil, err
-	}
-	row["addresses"] = addresses
+	row["addresses"] = strings.Join(addr, " ")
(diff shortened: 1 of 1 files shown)


trozet:
@abhat pointed out in the OVN NB docs it says:
       Addressing:
         addresses                   set of strings
         dynamic_addresses           optional string
         port_security               set of strings

So the doc says it should take a set, but doesn't seem to work for us.

abhat:
The [schema](http://www.openvswitch.org/support/dist-docs/ovn-nb.5.html) says that addresses is a set of strings. By conflating all the address in a single slice entry is a work-around, but I think the implementation should be conforming to the spec than work-around the existing ovn-northd limitation. We need clarification if this is a northd issue or if the spec is misplaced in the description of the addresses field.

trozet:
After some discussion it looks like the OVN definition of an address element in the set should be ("mac ip", "mac2 ip2), not ("mac", "ip", "mac2", "ip2"). This should be clarified in the OVN documentation. Don't think there is a need for this PR anymore.

PR #95, opened 2020-08-07
add support for chassisList API

girishmg:
we have a need to list all the chassis from the OVN SB. this commit
adds support for it and also updates the test for the same

@hzhou8 @vtolstov @noah8713 PTAL

### DIFF
--- a/chassis.go
+++ b/chassis.go
@@ -123,6 +123,28 @@ func (odbi *ovndb) chassisDelImp(name string) (*OvnCommand, error) {
 	return &OvnCommand{operations, odbi, make([][]map[string]interface{}, len(operations))}, nil
 }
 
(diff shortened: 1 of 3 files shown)


chassis.go
```diff
@@ -123,6 +123,29 @@ func (odbi *ovndb) chassisDelImp(name string) (*OvnCommand, error) {
...
+		return nil, ErrorSchema
+	}
+
+	for uuid := range cacheChassis {
+		ch, err := odbi.rowToChassis(uuid)
+		if err != nil {
+			return nil, err
+		}
+		listChassis = append(listChassis, ch)
+		fmt.Printf("uuid is %v\n", uuid)
```
noah8713, line 144: nit: no need to print uuid.

girishmg: @noah8713: Good catch. That was left over from some testing. I have fixed it now

PR #96, opened 2020-08-07
handle error in operations' result from transact method correctly

girishmg:
the results of the operations specified in the transact method is not
correctly verified. we need to always walk through the operations'
results to check for null error to claim a transaction has succeded.

furthermore, per RFC 7047 section 4.1.3, if all of the operations
succeed, but the results cannot be committed, then "result" will have
one more element than "params", with the additional element being an
<error>. this case was not handled as well.

with the fix in place, errors related to "constraint violation",
 "referential integrity violation", "resources exhausted", and
 "I/O error" are being caught.

### DIFF
--- a/ovnimp.go
+++ b/ovnimp.go
@@ -25,6 +25,10 @@ import (
 	"github.com/ebay/libovsdb"
 )
 
+const (
+	commitTransactionText = "commiting transaction"
+)
+
 var (
 	// ErrorOption used when invalid args specified
 	ErrorOption = errors.New("invalid option specified")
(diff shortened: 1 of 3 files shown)


ovnimp.go
```diff
@@ -149,12 +149,23 @@ func (odbi *ovndb) transact(db string, ops ...libovsdb.Operation) ([]libovsdb.Op
...
-				return nil, fmt.Errorf("Transaction Failed due to an error : %v details: %v in %v", o.Error, o.Details, ops[i])
+	// Per RFC 7047 Section 4.1.3, the operation result array in the transact response object
+	// maps one-to-one with operations array in the transact request object. We need to check
+	// each of the operation result for null error to ensure that the transaction has succeeded.
+	for i, o := range reply {
+		if o.Error != "" {
+			// Per RFC 7047 Section 4.1.3, if all of the operations succeed, but the results
+			// cannot be committed, then "result" will have one more element than "params",
+			// with the additional element being an <error>.
+			opsInfo := "commiting transaction"
```
noah8713, line 160: Move this to constant either in this file or common.go

girishmg: @noah8713 done. PTAL

vtolstov:
LGTM

PR #98, opened 2020-08-08
Support for monitoring selected tables in OVN NB or SB DB

girishmg:
this commit adds a new field to Config struct called TableCols which
is of type map[string][]string (maps table names to columns). with this
user can specify what tables in NB or in SB they would like to Monitor.

in a large scale OVN K8s cluster, the SB DB will be huge with 1000s of
entries in various tables (Logical_Flow, Port_Binding, adn so on),
neither do we want to get the initial data for these tables nor we want
updates for these tables and this commit provides a way to specify the
tables the user is interested in

adding support for specific columns within a table needs some more work
since all of the rowTo<tableName>() functions need to be updated to
reflect missing fields in the Row (since client expressed interest in
only subset of rows)

@noah8713 @hzhou8 @vtolstov PTAL

### DIFF
--- a/client.go
+++ b/client.go
@@ -219,6 +219,7 @@ type ovndb struct {
 	disconnectCB OVNDisconnectedCallback
 	db           string
 	addr         string
+	tableCols    map[string][]string
 	tlsConfig    *tls.Config
 	reconn       bool
 }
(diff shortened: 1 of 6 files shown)


vtolstov:
LGTM

PR #99, opened 2020-08-08
use 'ovs-appctl exit' instead of 'pkill ovsdb-server'

girishmg:
the current invocation of `pkill ovsdb-server` kills all instances of
ovsdb-server on the build machine. my build machine uses OVS on the
host for switching and it kills the host's ovsdb-server instance as well


vtolstov:
Thanks

PR #100, opened 2020-08-10
Make table constants as exportable

noah8713:
Since clients can watch on specific nb/sb tables, client can re-use
those table constants from go-ovn directly.
Hence, using upper case for tables to make those identifiers exportable.


noah8713:
@jxiaobin @hzhou8 @girishmg  PTAL

girishmg:
@noah8713 LGTM

vtolstov:
Thank you

Issue #102, opened 2020-08-14
In all of *ListImp() APIs allocate a slice with known capacity 


girishmg:
At scale, we are allocating more memory than that is required. Since we already know the number of logical resources ahead of time, just allocate the slice with the correct capacity.

For example, lsListImp() is better implemented as below.

```
func (odbi *ovndb) lsListImp() ([]*LogicalSwitch, error) {
	odbi.cachemutex.RLock()
	defer odbi.cachemutex.RUnlock()

	cacheLogicalSwitch, ok := odbi.cache[TableLogicalSwitch]
	if !ok {
		return nil, ErrorSchema
	}

	listLS := make([]*LogicalSwitch, len(cacheLogicalSwitch))
	i := 0
	for uuid := range cacheLogicalSwitch {
		listLS[i] = odbi.rowToLogicalSwitch(uuid)
		i++
	}

	return listLS, nil
}
```

cathy-zhou:
will take a look

PR #103, opened 2020-08-14
allocate slice with known capacity in *ListImpl() functions

cathy-zhou:
The current implementation of *ListImpl() functions allocate more memory
than that is required. We can just allocate the slice with exact
capacity as we know the number of the logical resources.

Fixes issue (#102)

Signed-off-by: Yun Zhou <yunz@nvidia.com>

### DIFF
--- a/acl.go
+++ b/acl.go
@@ -300,8 +300,6 @@ func (odbi *ovndb) rowToACL(uuid string) *ACL {
 
 // Get all acl by lswitch
 func (odbi *ovndb) aclListImp(lsw string) ([]*ACL, error) {
-	var listACL []*ACL
-
 	odbi.cachemutex.RLock()
 	defer odbi.cachemutex.RUnlock()
 
@@ -309,7 +307,6 @@ func (odbi *ovndb) aclListImp(lsw string) ([]*ACL, error) {
 	if !ok {
 		return nil, ErrorNotFound
 	}
-	var lsFound bool
 	for _, drows := range cacheLogicalSwitch {
 		if rlsw, ok := drows.Fields["name"].(string); ok && rlsw == lsw {
 			acls := drows.Fields["acls"]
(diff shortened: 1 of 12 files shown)


acl.go
```diff
@@ -300,43 +301,44 @@ func (odbi *ovndb) rowToACL(uuid string) *ACL {
...
-	var lsFound bool
 	for _, drows := range cacheLogicalSwitch {
 		if rlsw, ok := drows.Fields["name"].(string); ok && rlsw == lsw {
 			acls := drows.Fields["acls"]
 			if acls != nil {
 				switch acls.(type) {
 				case libovsdb.OvsSet:
 					if as, ok := acls.(libovsdb.OvsSet); ok {
-						for _, a := range as.GoSet {
+						listACL := make([]*ACL, len(as.GoSet))
```
noah8713, line 318: Thanks for optimization and improvements.

Instead of  manually tracking a new index i, better set capacity for slice as length is known which will save unnecessary memory allocations too
e.g. listACL := make([]*ACL, 0, len(as.GoSet)). 

keep the listACL = append(listACL, ta) as is so that it will add from start of slice.

cathy-zhou: @noah8713 Thanks. I thought append() function always reallocate memory for the new result, apprently it also reuse the memory as long as the capacity is sufficient. I will update my code and repush. Thanks again.

acl.go
```diff
@@ -300,43 +301,44 @@ func (odbi *ovndb) rowToACL(uuid string) *ACL {
...
+						for i, a := range as.GoSet {
 							if va, ok := a.(libovsdb.UUID); ok {
 								ta := odbi.rowToACL(va.GoUUID)
-								listACL = append(listACL, ta)
+								listACL[i] = ta
+							} else {
```
noah8713, line 323: @cathy-zhou  Also, one more thing. A separate PR/commit for error handling would be preferred. Hence, this patch will be just to initialize slices with proper mem allocations

cathy-zhou: sure, thanks for your comments. I just repushed the updated code.

logical_router.go
```diff
@@ -330,20 +329,16 @@ func (odbi *ovndb) lrlbListImp(lr string) ([]*LoadBalancer, error) {
...
 					} else {
 						return nil, fmt.Errorf("type libovsdb.UUID casting failed")
 					}
 				default:
 					return nil, fmt.Errorf("Unsupport type found in ovsdb rows")
 				}
 			}
-			lrFound = true
-			break
+			return []*LoadBalancer{}, nil
```
vtolstov, line 340: this breaks older behaviour - before changes if list return empty slice it returns ErrorNotFound

cathy-zhou: @vtolstov Can you please clarify? In the old code, if lrFound is true, it returns empty list. It only returns ErrorNotFound if the specific logical router is not found, which is the case in the new code as well.

noah8713: I revisited this and it seems @vtolstov  is right. We can't just return because for each of the entities in for loop, there are two cases, ovsset and uuid. Hence, static slice capacity will not help here. We need to use dynamic slice  which is len(go.set)+ uuids for which we don't know how many are in place. We return entire slice accordingly . Hence, (keep the current code as is). What do you think? Please feel free to extend a UT for the same.

cathy-zhou: @noah8713 Here is my understanding: that for loop is only used to find the right router, and if it finds the right router, it breaks. We don't continue the for loop to find more entries. So once the right router is found, there are only two cases: ovsset or uuid. In both cases, we can preallocate the right size of slice.

Please let me know if my understanding is incorrect. Thanks!

noah8713: Ah I see, I guess I got little confused again reading size as uuid + go.set as I saw @vtolstov comment to see if I didn't miss anything. LGTM again :)

cathy-zhou: thanks!

PR #105, opened 2020-08-27
GoSet could have not just strings but also set of uuids

girishmg:
This PR fixes following issues:

- fix logical_router_static_route table's column type assumption

  the code currently assumes that the policy and output_port columns of
logical_router_static_route table to be of type UUID, however it is
of type string

- GoSet could have not just strings but also set of uuids

   the current ConvertGoSetToStringArray() API assumes that the GoSet
consists of only strings but it could also contain set of UUIDs (for
example logical_switch`ports column).

### DIFF
--- a/ovnimp.go
+++ b/ovnimp.go
@@ -334,9 +334,13 @@ func (odbi *ovndb) populateCache(updates libovsdb.TableUpdates) {
 func (odbi *ovndb) ConvertGoSetToStringArray(oset libovsdb.OvsSet) []string {
 	var ret = []string{}
 	for _, s := range oset.GoSet {
-		value, ok := s.(string)
-		if ok {
+		switch s.(type) {
+		case string:
+			value := s.(string)
 			ret = append(ret, value)
+		case libovsdb.UUID:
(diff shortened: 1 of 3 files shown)


girishmg:

vtolstov (approved):

ovnimp.go
```diff
@@ -334,9 +334,13 @@ func (odbi *ovndb) populateCache(updates libovsdb.TableUpdates) {
 func (odbi *ovndb) ConvertGoSetToStringArray(oset libovsdb.OvsSet) []string {
 	var ret = []string{}
 	for _, s := range oset.GoSet {
-		value, ok := s.(string)
-		if ok {
+		switch s.(type) {
+		case string:
+			value := s.(string)
 			ret = append(ret, value)
+		case libovsdb.UUID:
```
noah8713, line 341: Thanks for the improvement. Seems we should add at-least one UT to cover this so that uuids are also included as ConvertGoSetToStringArray is common module uses by all entities.

girishmg: @noah8713 i have added the Unit Testcase, so PTAL.

Issue #107, opened 2020-08-27
Update testing environment 


JacobTanenbaum:
The testing environment is out of date, It would be better to bring in a more updated version of openvswitch and ovn to test against. 

As part of updating LRSRAdd() to support ecmp routes it I found that the version of ovs/ovn tested against does not support ecmp routes.

noah8713:
yes ci can be improved to be switched to use ovs master and ovn 20.xx (latest)/master. Do you want to pick it up ? Else we can plan to pick it too cc @vtolstov @hzhou8

JacobTanenbaum:
Yes, I should be able to pick this up

PR #108, opened 2020-08-28
updated the testing to use a more updated ovs/ovn environment

JacobTanenbaum:
Signed-off-by: Jacob Tanenbaum <jtanenba@redhat.com>

addresses issue: https://github.com/eBay/go-ovn/issues/107

### DIFF
--- a/.travis/test.sh
+++ b/.travis/test.sh
@@ -1,7 +1,10 @@
 #!/bin/sh -ex
 
 rm -rf ovs
-git clone --depth 1 -b branch-2.12 https://github.com/openvswitch/ovs.git
+rm -rf ovn
+
+git clone --depth 1 -b master https://github.com/openvswitch/ovs.git
+git clone --depth 1 -b master https://github.com/ovn-org/ovn.git
 cd ovs
 ./boot.sh && ./configure --enable-silent-rules
 make -j4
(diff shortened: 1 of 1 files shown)


.travis/test.sh
```diff
@@ -1,7 +1,10 @@
 #!/bin/sh -ex
 
 rm -rf ovs
-git clone --depth 1 -b branch-2.12 https://github.com/openvswitch/ovs.git
+rm -rf ovn
+
+git clone --depth 1 -b master https://github.com/openvswitch/ovs.git
+git clone --depth 1 -b  branch-20.06 https://github.com/ovn-org/ovn.git
```
noah8713, line 7: Thanks for taking this up. Can we use master for ovn too because pinning to 20.06 will again result in new feature miss? Also, you can refer to sample build script to configure ovs/ovn seperately as source referring to https://github.com/ovn-org/ovn/blob/master/.travis/linux-build.sh .

JacobTanenbaum: I updated to pull ovn from master.

I am not sure why splitting ovn and ovs setup would be beneficial... There is no real setup for ovn while setting up ovs the database schema is taken from the ovn repo to apply to the ovs dbs.

noah8713: Thanks. Yeah that is also valid point as for sandbox we don't really need all of it.

vtolstov (approved):

PR #109, opened 2020-10-02
Fix LogicalRouterStaticRoute related APIs

hzhou8:
The APIs LRSRAdd and LRSRDel were broken when output_port and policy parameters were used. Test case was missing this part. The first patches in the PR fix the bug. The latter patches update the API signature to be more straightforward. Although the API signature is changed, it shouldn't create compatibility problem given that the APIs were broken before this PR.


jxiaobin (commented):
lgtm

softwarejl (approved):
LGTM

PR #110, opened 2020-10-02
Refactor getRowUUIDs()

hzhou8:
The current logic is correct but a little confusing. The flag is initialized
to false, and flips to true if the *first* column matches which is unnecessary.

Signed-off-by: Han Zhou <hzhou8@ebay.com>


jxiaobin (commented):
lgtm

softwarejl (approved):
LGTM

PR #111, opened 2020-10-07
Refactor rowToLogicalRouter() and RowToPortGroup()

noah8713:
To be consistent and make the code more neat instead of typing
same table name with uuid column.

Signed-off-by: Aliasgar Ginwala <aginwala@ebay.com>


vtolstov (approved):
LGTM

Issue #112, opened 2020-10-29
Add new loadbalancer fields


aojea:
The OVN LoadBalancer added some new fields in the last releases
```
   Summary:
       name                          string
       vips                          map of string-string pairs
       protocol                      optional string, one of sctp, tcp, or udp
       health_check                  set of Load_Balancer_Health_Checks
       ip_port_mappings              map of string-string pairs
       selection_fields              set of strings, one of  eth_dst,  eth_src,  ip_dst,  ip_src,
                                     tp_dst, or tp_src
       Common Columns:
         external_ids                map of string-string pairs
```

but go-ovn only have

https://github.com/eBay/go-ovn/blob/da67e9744ec05bf817ef2de7128971f235b921c3/load_balancer.go#L25-L32

xref:
https://github.com/ovn-org/ovn/blob/master/ovn-nb.ovsschema#L188
http://manpages.ubuntu.com/manpages/hirsute/man5/ovn-nb.5.html#load_balancer%20table

aojea:
I can work on this, just need some guidance on how to implement new fields keeping compatibility between ovn versions

hzhou8:
@aojea Thanks for volunteering on this. For compatibility:
1) from OVSDB to go-ovn, if a column doesn't exist, just leave the go-ovn fields empty.
2) from go-ovn to OVSDB, if a column doesn't exist in schema, then don't add it to the transaction row. This may need a little more effort to check the DB schema.

PR #113, opened 2020-11-01
add selection_fields to the load balancer 

aojea:
Since OVN version 20.06.0, load balancers have a new column
'selection_fields' that allows to define a set of packet
headers to use while selecting the backend, so that session
affinity can be implemented.

Signed-off-by: Antonio Ojea <aojea@redhat.com>

Fixes: https://github.com/eBay/go-ovn/issues/112

### DIFF
--- a/load_balancer.go
+++ b/load_balancer.go
@@ -24,11 +24,12 @@ import (
 
 // LoadBalancer ovnnb item
 type LoadBalancer struct {
-	UUID       string
-	Name       string
-	vips       map[interface{}]interface{}
-	protocol   string
-	ExternalID map[interface{}]interface{}
+	UUID            string
+	Name            string
+	vips            map[interface{}]interface{}
+	protocol        string
+	selectionFields string
+	ExternalID      map[interface{}]interface{}
 }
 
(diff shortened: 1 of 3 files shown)


aojea:
/cc @dcbw 
@hzhou8 is still WIP because I need to verify the tests work correctly, but can you please take a look and provide feedback?

load_balancer.go
```diff
@@ -158,6 +159,22 @@ func (odbi *ovndb) lbGetImp(name string) ([]*LoadBalancer, error) {
 	return listLB, nil
 }
 
+func (odbi *ovndb) lbSetSelectionFieldsImp(name string, selectionFields string) (*OvnCommand, error) {
+	row := make(OVNRow)
+	row["selection_fields"] = selectionFields
+
+	condition := libovsdb.NewCondition("name", "==", name)
+
+	insertOp := libovsdb.Operation{
```
hzhou8, line 168: The variable name is misleading. It is better to be updateOp.

aojea: thanks

hzhou8 (approved):

softwarejl (approved):

PR #114, opened 2020-11-20
ovnsignal: expose all methods as public

zhouhaibing089:
Fixes the OVNSignal interface to let its all methods to be public. This
is necessary from consumer side to define OVNSignal implementation.


vtolstov (approved):
LGTM

hzhou8 (approved):

PR #117, opened 2020-12-11
Port group API support

anfredette:
ovn-kubernetes needs the ability to add ACLs to port groups, but the current ACL APIs only support logical switches, and there are no port group APIs. This PR adds port group API support and tests.

### DIFF
--- a/client.go
+++ b/client.go
@@ -206,6 +206,15 @@ type Client interface {
 	// Get SB_Global table options
 	SBGlobalGetOptions() (map[string]string, error)
 
+	// Creates a new port group in the Port_Group table named "group" with optional "ports" added to the group.
+	PortGroupAdd(group string, ports []string, external_ids map[string]string) (*OvnCommand, error)
+	// Sets "ports" and/or "external_ids" on the port group named "group". It is an error if group does not exist.
+	PortGroupUpdate(group string, ports []string, external_ids map[string]string) (*OvnCommand, error)
+	// Deletes port group "group". It is an error if "group" does not exist.
+	PortGroupDel(group string) (*OvnCommand, error)
+	// Get PortGroup data structure if it exists
+	PortGroupGet(group string) (*PortGroup, error)
+
 	// Close connection to OVN
 	Close() error
 }
@@ -667,6 +676,22 @@ func (c *ovndb) SBGlobalSetOptions(options map[string]string) (*OvnCommand, erro
 
 func (c *ovndb) SBGlobalGetOptions() (map[string]string, error) {
 	return c.sbGlobalGetOptionsImp()
+}
+
+func (c *ovndb) PortGroupAdd(group string, ports []string, external_ids map[string]string) (*OvnCommand, error) {
+	return c.pgAddImp(group, ports, external_ids)
+}
+
+func (c *ovndb) PortGroupUpdate(group string, ports []string, external_ids map[string]string) (*OvnCommand, error) {
+	return c.pgUpdateImp(group, ports, external_ids)
+}
+
+func (c *ovndb) PortGroupDel(group string) (*OvnCommand, error) {
+	return c.pgDelImp(group)
+}
+
+func (c *ovndb) PortGroupGet(group string) (*PortGroup, error) {
+	return c.pgGetImp(group)
 }
 
 // these functions are helpers for unit-tests, but not part of the API
--- a/port_group.go
+++ b/port_group.go
@@ -43,11 +43,19 @@ func (odbi *ovndb) pgAddImp(group string, ports []string, external_ids map[strin
 	if uuid := odbi.getRowUUID(TablePortGroup, row); len(uuid) > 0 {
 		return nil, ErrorExist
 	}
-	pgports, err := libovsdb.NewOvsSet(ports)
-	if err != nil {
-		return nil, err
+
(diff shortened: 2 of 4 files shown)


anfredette:
@hzhou8 @abhat @vjayaramrh please review.

vtolstov (approved):
LGTM

client.go
```diff
@@ -206,6 +206,15 @@ type Client interface {
...
 	SBGlobalGetOptions() (map[string]string, error)
 
+	// Creates a new port group in the Port_Group table named "group" with optional "ports" added to the group.
+	PortGroupAdd(group string, ports []string, external_ids map[string]string) (*OvnCommand, error)
+	// Sets "ports" and/or "external_ids" on the port group named "group". It is an error if group does not exist.
+	PortGroupSet(group string, ports []string, external_ids map[string]string) (*OvnCommand, error)
+	// Deletes port group "group". It is an error if "group" does not exist.
+	PortGroupDel(group string) (*OvnCommand, error)
+	// Get PortGroup data structure if it exists
+	PortGroupGet(group string) ([]*PortGroup, error)
```
hzhou8, line 216: Port_Group's name column is unique (index), so I think it doesn't make sense to return an array. It is better to return a pointer - nil if doesn't exist.

client.go
```diff
@@ -206,6 +206,15 @@ type Client interface {
 	// Get SB_Global table options
 	SBGlobalGetOptions() (map[string]string, error)
 
+	// Creates a new port group in the Port_Group table named "group" with optional "ports" added to the group.
+	PortGroupAdd(group string, ports []string, external_ids map[string]string) (*OvnCommand, error)
+	// Sets "ports" and/or "external_ids" on the port group named "group". It is an error if group does not exist.
+	PortGroupSet(group string, ports []string, external_ids map[string]string) (*OvnCommand, error)
```
hzhou8, line 212: For most APIs of this library, the naming convention is "Update" for the whole row, "Set" for a specific column. Since this API updates multiple columns, it may be better named as PortGroupUpdate(...).

In addition, there is a column ACL. Would you want to include in this API so that it can be updated? Same for PortGroupAdd() API.

port_group_test.go
```diff
@@ -17,256 +17,367 @@
...
-	PG_TEST_EXT_ID_MAC     = "00:01:02:03:04:05"
-	PG_TEST_EXT_ID_MAC_2   = "01:02:03:05:05:06"
-	PG_TEST_EXT_ID_IP_KEY  = "ip_addr"
-	PG_TEST_EXT_ID_IP      = "169.254.1.1"
-	PG_TEST_OPT_1_KEY      = "foo1"
-	PG_TEST_OPT_1_VAL      = "bar1"
-	PG_TEST_OPT_1_VAL_2    = "baz1"
-	PG_TEST_OPT_2_KEY      = "foo2"
-	PG_TEST_OPT_2_VAL      = "bar2"
+	PG_TEST_PG1         = "TestPortGroup1"
```
hzhou8, line 29: This commit seems to be a fix/revision of the previous one. I think it is better to be combined.

hzhou8 (changes requested):
Thanks @anfredette for helping on this. Please see my comments.

client.go
anfredette: Will do.

client.go
anfredette: I'll update the name. I was intending to just add the API support for the existing port-group functionality in this PR, and then add support for the ACL column when I added the generic ACL support that we discussed so I could test them together.  Is this okay with you?

port_group_test.go
anfredette: Yup. I'll squash them.

anfredette:
Vasily and Han, thanks for the review!

anfredette:
I think I've addressed all of the comments, except for adding support for the ACL column which I plan to take care of when I add ACL support for port-groups.

port_group.go
```diff
@@ -43,11 +43,19 @@ func (odbi *ovndb) pgAddImp(group string, ports []string, external_ids map[strin
 	if uuid := odbi.getRowUUID(TablePortGroup, row); len(uuid) > 0 {
 		return nil, ErrorExist
 	}
-	pgports, err := libovsdb.NewOvsSet(ports)
-	if err != nil {
-		return nil, err
+
+	if ports != nil {
+		var portUUIDs []libovsdb.UUID
```
vtolstov, line 48: portUUIDs := make([]libovsdb.UUID, 0, len(ports))

port_group.go
```diff
@@ -66,18 +74,30 @@ func (odbi *ovndb) pgAddImp(group string, ports []string, external_ids map[strin
...
-	if err != nil {
-		return nil, err
+
+	if ports == nil && external_ids == nil {
+		return nil, ErrorNoChanges
 	}
-	row["ports"] = pgports
+
+	if ports != nil {
+		var portUUIDs []libovsdb.UUID
```
vtolstov, line 90: portUUIDs := make([]libovsdb.UUID, 0, len(ports))

vtolstov (commented):
minor comments about optimizations, lgtm

anfredette:
@vtolstov, thanks for the suggestion.  I'll make that optimization in my next PR.

PR #118, opened 2020-12-15
use build tag to group tests that needs to be mandatorily run as part of travis test script

vjayaramrh:
This PR adds a build tag `travis` for the files `sb_global_test.go` and `nb_global_test.go`. The tests `TestSBGlobalAPI` and `TestNBGlobalAPI`run successfully only when the `.travis\test.sh` script is invoked which is long running and is needed as part of travis CI run.
For all other tests, the regular go testing command 'go test' should be sufficent and faster. So now with this PR, one can execute `OVN_NB_DB=unix:ovnnb_db.sock OVN_SB_DB=unix:ovnsb_db.sock OVS_RUNDIR=/var/run/ovn go test` which will skip the TestSBGlobalAPI` and `TestNBGlobalAPI` tests but run everything else. The TestSBGlobalAPI` and `TestNBGlobalAPI` tests will still continue to be run along with the other tests when the `.travis/test.sh` is executed

Signed-off-by: Vishwanath Jayaraman <vjayaram@redhat.com>

### DIFF
--- a/.travis/test.sh
+++ b/.travis/test.sh
@@ -112,7 +112,7 @@ ovn_start_db sb standalone 1 $srcdir/../ovn/ovn-sb.ovsschema
 export GO111MODULE=on
 cd ../
 go get -v ./...
-go test -v
+go test -v -tags=travis
 
 ovs-appctl -t ${OVS_RUNDIR}/nb1 exit
 ovs-appctl -t ${OVS_RUNDIR}/sb1 exit
--- a/nb_global_test.go
+++ b/nb_global_test.go
@@ -1,3 +1,5 @@
+// +build travis
+
 /**
  * Copyright (c) 2020 eBay Inc.
  *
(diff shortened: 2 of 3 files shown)


hzhou8:
LGTM

hzhou8 (approved):

vtolstov:
LGTM

vtolstov (approved):

anfredette:
Per my comments in Slack, another approach is to fix TestNBGlobalAPI and TestSBGlobalAPI so that they check whether an entry has already been created before trying to create another one which is an action that is guaranteed to fail.  This way the tests will run whether the databases have been initialized or not.  So, if we want to explicitly test the internal nb/sb Add functions, the test can be run on uninitialized databases, for example with the .travis/test.sh test script.

anfredette (approved):
After talking with Vish, I think this approach is preferable to the one I suggested.  In particular, with the build tag approach, the user needs to explicitly run these test cases, and if they want them to pass, they need to ensure that the global table entries have not been created before running them.

LGTM

Issue #119, opened 2020-12-18
crash on populateCache


amorenoz:
Sometimes, when the cache is populated (e.g: on `goovn.NewClient()`) the library _can_ crash the program (in several ways). An example crash traceback is here:

```
panic: interface conversion: interface {} is nil, not string
goroutine 94 [running]:
github.com/ebay/go-ovn.(*ovndb).rowToLogicalRouterPort(0xc000e89c80, 0xc0024bdcb0, 0x24, 0x5)
        /home/amorenoz/src/skynet/workdir/go/pkg/mod/github.com/amorenoz/go-ovn@v0.1.1-0.20201111124657-2a0d4d618412/logical_router_port.go:136 +0x14c4
github.com/ebay/go-ovn.(*ovndb).rowToLogicalRouter(0xc000e89c80, 0xc0024bdd40, 0x24, 0x24)
        /home/amorenoz/src/skynet/workdir/go/pkg/mod/github.com/amorenoz/go-ovn@v0.1.1-0.20201111124657-2a0d4d618412/logical_router.go:159 +0x1539
github.com/ebay/go-ovn.(*ovndb).populateCache(0xc000e89c80, 0xc0025daa50)
        /home/amorenoz/src/skynet/workdir/go/pkg/mod/github.com/amorenoz/go-ovn@v0.1.1-0.20201111124657-2a0d4d618412/ovnimp.go:235 +0x24d5
github.com/ebay/go-ovn.connect(0xc000e89c80, 0x0, 0x0)
        /home/amorenoz/src/skynet/workdir/go/pkg/mod/github.com/amorenoz/go-ovn@v0.1.1-0.20201111124657-2a0d4d618412/client.go:261 +0x1f7
github.com/ebay/go-ovn.NewClient(0xc0008ede70, 0xc002594600, 0x1f, 0x0, 0x0)
        /home/amorenoz/src/skynet/workdir/go/pkg/mod/github.com/amorenoz/go-ovn@v0.1.1-0.20201111124657-2a0d4d618412/client.go:290 +0x39f
github.com/skydive-project/skydive/topology/probes/ovn.(*Probe).Do(0xc000170000, 0x7c23700, 0xc0016e0600, 0xc00044a820, 0x0, 0xc000743ec0)
        /home/amorenoz/src/skynet/workdir/go/src/github.com/skydive-project/skydive/topology/probes/ovn/ovn.go:714 +0x228
[...]

PR #120, opened 2020-12-18
onvimp: enforce table ordering in cache population

amorenoz:
When the cache is populated, a specific order must be preserved that
ensures that if table A refereces table B, B is populated before A.

This assumption seems widely made in the RowTo{XXX} functions. So,
this patch enforces this ordering.

Fixes: #119 
Signed-off-by: Adrian Moreno <amorenoz@redhat.com>

### DIFF
--- a/common.go
+++ b/common.go
@@ -54,6 +54,10 @@ const (
 	TableSBGlobal                 string = "SB_Global"
 )
 
+/* Ordered NorthBound Tables
+The cache will be populated following this order.
+Care must be taken too keep reference targets before their sources.
+*/
 var NBTablesOrder = []string{
 	TableNBGlobal,
 	TableAddressSet,
@@ -76,6 +80,10 @@ var NBTablesOrder = []string{
 	TableLogicalRouter,
 }
 
+/* Ordered SouthBound Tables
+The cache will be populated following this order.
+Care must be taken too keep reference targets before their sources.
+*/
 var SBTablesOrder = []string{
 	TableChassis,
 	TableEncap,
--- a/ovnimp.go
+++ b/ovnimp.go
@@ -210,7 +210,17 @@ func (odbi *ovndb) populateCache(updates libovsdb.TableUpdates) {
 	odbi.cachemutex.Lock()
 	defer odbi.cachemutex.Unlock()
 
-	for table := range odbi.tableCols {
+	var tableOrder []string
+	if odbi.db == DBNB {
+		tableOrder = NBTablesOrder
+	} else {
+		tableOrder = SBTablesOrder
+	}
+
+	for _, table := range tableOrder {
+		if _, ok := odbi.tableCols[table]; !ok {
+			continue
+		}
 		tableUpdate, ok := updates.Updates[table]
 		if !ok {
 			continue


hzhou8:
Thanks for the patch. Could you help give an example where a RowToXXX function depends on the ordering? I am not aware of such kind of dependency except one that I fixed/removed in an previous commit.

amorenoz:
Hi @hzhou8, sorry I did not put the <Fixes> link on the original description (fixed that now).
As you can see on backtrace pasted in the linked Issue (#119), the RowToLogicalRouter depends on the on the LogicalRouterPort cache being populated.

However, I've just realized I was testing without https://github.com/eBay/go-ovn/commit/5066560cee670596aab3c3da959e01b160953431. That patch may have solved my issue. I'll retest.

OTOH, I think the intended behaviour of `populateCache` was to respect the order of the `tablesOrder` list. This behaviour was changed with https://github.com/eBay/go-ovn/commit/30abed5fb9684c24dd9b93f7a5000a85aeae6a1f. So restoring that enforcement might be a good idea even the inter-table dependency has been fixed.

amorenoz:
@hzhou8 , indeed https://github.com/eBay/go-ovn/commit/5066560cee670596aab3c3da959e01b160953431 does fix my issue.
Feel free to close this PR or consider the last argument stated in my previous comment.

hzhou8:
@amorenoz Thanks for the context. In fact I don't think the ordering is necessary, at least for now. In the current design it should not assume any dependency while populating cache, because the references are stored as uuids only. Thank you for pointing out that there was a behavior change introduced by https://github.com/eBay/go-ovn/commit/30abed5fb9684c24dd9b93f7a5000a85aeae6a1f, but after going deeper in the history it seems that the tablesOrder introduced by commit 30abed5 was intended only to fix this same problem of commit 82a71f5067. If commit 82a71f5067 didn't try to resolve the references, this whole tablesOrder logic wasn't necessary. I am sorry that all these slipped from the code reviews in the first place. So I would rather keep it as is, or even better, to remove the tablesOrder mechanism, to avoid introducing more confusion. What do you think?

amorenoz:
> @amorenoz Thanks for the context. In fact I don't think the ordering is necessary, at least for now. In the current design it should not assume any dependency while populating cache, because the references are stored as uuids only. Thank you for pointing out that there was a behavior change introduced by [30abed5](https://github.com/eBay/go-ovn/commit/30abed5fb9684c24dd9b93f7a5000a85aeae6a1f), but after going deeper in the history it seems that the tablesOrder introduced by commit [30abed5](https://github.com/eBay/go-ovn/commit/30abed5fb9684c24dd9b93f7a5000a85aeae6a1f) was intended only to fix this same problem of commit [82a71f5](https://github.com/eBay/go-ovn/commit/82a71f5067fb1bb7bdb357d35cadec656872cc31). If commit [82a71f5](https://github.com/eBay/go-ovn/commit/82a71f5067fb1bb7bdb357d35cadec656872cc31) didn't try to resolve the references, this whole tablesOrder logic wasn't necessary. I am sorry that all these slipped from the code reviews in the first place. So I would rather keep it as is, or even better, to remove the tablesOrder mechanism, to avoid introducing more confusion. What do you think?

Agree. It's definitely more robust not to depend on a specific ordering.

amorenoz:
Closing. Sorry for the noise.

PR #121, opened 2020-12-21
Facilitate local unit tests

amorenoz:
While doing some changes to go-ovn I wanted to have a quick and isolated way to run the go-ovn unit tests (instead of relying on travis or having to do a lot of preliminary setup). So I ended up reusing the travis scripts to have a container-isolated local `Makefile`.

The result is that, running `make check` will build a container (only the first time you run it) and run the unit tests using the same scripts travis.

Sending this PR as a request for comments. Is something like this considered useful?


hzhou8:
Thanks @amorenoz .  Yes, this is looks very reasonable and useful. What's the work pending (I am asking since it is marked as RFC)?

amorenoz:
@hzhou8 I marked as RFC because I was not sure if it would be considered useful. If it is, I think it's ready for review

.travis/test_prepare.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/sh -ex
+
+SCRIPT=`realpath -s $0`
+SCRIPTPATH=`dirname $SCRIPT`
+
+. $SCRIPTPATH/test_common.sh
+
+(
+echo "building ovn is ${ovn_srcdir}"
```
hzhou8, line 9: typo: s/is/in

.travis/test_run.sh
```diff
@@ -0,0 +1,115 @@
...
+ovn_start_db nb standalone 1 ${ovn_srcdir}/ovn/ovn-nb.ovsschema
+ovn_start_db sb standalone 1 ${ovn_srcdir}/ovn/ovn-sb.ovsschema
+
+export GO111MODULE=on
+go get -v ./...
+go test -v -tags=travis
+
+ovs-appctl -t ${OVS_RUNDIR}/nb1 exit
+ovs-appctl -t ${OVS_RUNDIR}/sb1 exit
+
```
hzhou8, line 115: Need to delete db files in sandbox. Otherwise, there will be error when executing test_run.sh next time:

ovsdb-tool: I/O error: nb1.db: create failed (File exists)

Dockerfile
```diff
@@ -0,0 +1,22 @@
+from golang:1.12
+
+RUN apt-get update && apt-get install --no-install-recommends -y \
```
hzhou8, line 3: Not sure if it is my environment problem, but this always fails for me:

Step 1/6 : from golang:1.12
 ---> ffcaee6f7d8b
Step 2/6 : RUN apt-get update && apt-get install --no-install-recommends -y     bc     gcc-multilib     libssl-dev      llvm-dev     libjemalloc-dev     libnuma-dev     python-sphinx     libelf-dev     selinux-policy-dev     libunbound-dev     autoconf     automake     libtool
 ---> Running in 93a1daa2462e
Err:1 http://security.debian.org/debian-security buster/updates InRelease
  Temporary failure resolving 'security.debian.org'
Err:2 http://deb.debian.org/debian buster InRelease
  Temporary failure resolving 'deb.debian.org'
Err:3 http://deb.debian.org/debian buster-updates InRelease
  Temporary failure resolving 'deb.debian.org'
Reading package lists...
W: Failed to fetch http://deb.debian.org/debian/dists/buster/InRelease  Temporary failure resolving 'deb.debian.org'
W: Failed to fetch http://security.debian.org/debian-security/dists/buster/updates/InRelease  Temporary failure resolving 'security.debian.org'
W: Failed to fetch http://deb.debian.org/debian/dists/buster-updates/InRelease  Temporary failure resolving 'deb.debian.org'
W: Some index files failed to download. They have been ignored, or old ones used instead.

hzhou8 (changes requested):

Makefile
```diff
@@ -1,12 +1,18 @@
...
+DOCKER ?= $(shell which docker)
+IMAGE_NAME="goovn:test"
 
-# Remove all build artifacts.
-#
-# Example:
-#   make clean
+.PHONY=check
+check: check_prep
+	$(DOCKER) run -e "OVN_SRCDIR=/src/" -v $$PWD:/root/workspace -w /root/workspace -it $(IMAGE_NAME) .travis/test_run.sh
```
hzhou8, line 8: For the docker approach, "go get -v" is executed every time, but the result is lost after container exit. This step is also time-consuming. So I wonder if it is better to have it in test_prepare.sh as well. That way, the subsequent "make check" would be much faster.

.travis/test_run.sh
amorenoz: Since test_run.sh it is executed in a docker that is then removed, is it really necessary?
I don't think it hurts to delete the db files but I'd like to understand how have you tested it

Makefile
amorenoz: Maybe we run 'go get' in both scripts? That way live modifications of the go.mod file will be picked up while the go mod cache speeds things up. WDYT?

hzhou8: Yes, sounds good.

.travis/test_run.sh
hzhou8: I assume this can also be executed without using docker, if preferred. In that case the cleanup is helpful.

amorenoz: Yep. Makes sense

.travis/test_prepare.sh
amorenoz: Done

.travis/test_run.sh
amorenoz: Done

Makefile
amorenoz: Done

hzhou8:
LGTM

.travis/test_prepare.sh
hzhou8: Sorry, I didn't see it fixed in this revision. Will you send another revision?

.travis/test_run.sh
hzhou8: Sorry, I didn't see it changed in this revision. Will you send another revision?

hzhou8 (changes requested):

hzhou8:
LGTM now except the minor typo. @amorenoz let me know you want to push once again. Thanks a lot.

@vtolstov would you take a look?

vtolstov (approved):

vtolstov:
ok for me, minor typos in wording

.travis/test_prepare.sh
amorenoz: Sorry. I thought I'd fixed that one. I'll send another revision

vjayaramrh:
@amorenoz if I wanted to execute 'go test -v ./...' or run a single test such as 'go test -run TestLogicalRouterPort' (refer https://github.com/eBay/go-ovn/blob/master/logical_router_port_test.go#L10), how would I be able to do that? Thanks

hzhou8:
> @amorenoz if I wanted to execute 'go test -v ./...' or run a single test such as 'go test -run TestLogicalRouterPort' (refer https://github.com/eBay/go-ovn/blob/master/logical_router_port_test.go#L10), how would I be able to do that? Thanks

@vjayaramrh I think the test_run.sh script can be further improved to provide an option to run a specific test case. With this we can use "docker run ... sh" to start the container and execute the test_run.sh script for specific test cases.

I will merge this PR now since it would already provide lots of conveniences. More improvements are welcome :)

amorenoz:
@vjayaramrh, although it's possible to run your own "docker run" command, we could make it easier by adding an option, like @hzhou8 mentions, or even my documenting that process. WDYT?

PR #122, opened 2020-12-23
Added support for ACLs on port groups.

anfredette:
Added support for ACLs on port groups.
 - Using ovn-nbctl acl-add/del as a model, an entityType parameter was added to
the API which could be either PORT_GROUP or LOICAL_SWITCH.
- In order to maintain backward compatibility of the existing APIs, a new set
of APIs called ACLAddEntity/ACLDelEntity/ACLListEntity were added.
- Note: both versions of the APIs call the same underlying
implementation functions.
    
Also made a minor optimization suggested by vtolstov on pr-117.
    
### DIFF
--- a/acl.go
+++ b/acl.go
@@ -33,17 +33,28 @@ type ACL struct {
 	ExternalID map[interface{}]interface{}
 }
 
-func (odbi *ovndb) getACLUUIDByRow(lsw, table string, row OVNRow) (string, error) {
+func (odbi *ovndb) getACLUUIDByRow(entityType EntityType, entity string, row OVNRow) (string, error) {
 	odbi.cachemutex.RLock()
 	defer odbi.cachemutex.RUnlock()
 
-	cacheLogicalSwitch, ok := odbi.cache[TableLogicalSwitch]
+	var tableName string
+
+	switch entityType {
+	case LOGICAL_SWITCH:
+		tableName = TableLogicalSwitch
+	case PORT_GROUP:
+		tableName = TablePortGroup
+	default:
+		return "", ErrorOption
+	}
+
+	tableCache, ok := odbi.cache[tableName]
 	if !ok {
 		return "", ErrorSchema
 	}
 
-	for _, drows := range cacheLogicalSwitch {
-		if rlsw, ok := drows.Fields["name"].(string); ok && rlsw == lsw {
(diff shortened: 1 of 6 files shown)


hzhou8:
Thanks @anfredette . LGTM!

A minor comment is that it would be better to split the fixes such as for the problem when external_ids == nil in separate commits. Let me know if you'd like to do that.

anfredette:
I agree. I’ll split it out.

Thanks,
Andre

anfredette:
I've updated this PR as follows:
1. Created Issue #124 and split out fix as PR  #125.
2. Rebased this PR on PR #124.
3. Optimized getACLUUIDByRow() 
   - when we have the row UUID, we don't have to loop through the table to find it.

softwarejl (approved):
LGTM

acl.go
```diff
@@ -34,120 +34,160 @@ type ACL struct {
...
+func (odbi *ovndb) aclAddImp(entityType EntityType, entity, direct, match, action string, priority int, external_ids map[string]string, logflag bool, meter, severity string) (*OvnCommand, error) {
+	var table string
+	var condition []interface{}
+
+	switch entityType {
+	case LOGICAL_SWITCH:
+		condition = libovsdb.NewCondition("name", "==", entity)
+		table = TableLogicalSwitch
+	case PORT_GROUP:
+		condition = libovsdb.NewCondition("_uuid", "==", stringToGoUUID(entity))
```
hzhou8, line 185: Why using UUID for port_group (while we use "name" for logical_switch). Is it better to use "name" so that it is consistent between the two entity types?

anfredette:
Per Han's request above, and discussion on Slack, I've changed the port group ACL APIs to use the PG name instead of the PG UUID.

NOTE: I've kept this change as a separate commit to facilitate reviews.  After I get the thumbs up, I will squash the last two commits.

acl.go
anfredette: Done.

hzhou8 (approved):

hzhou8 (changes requested):
Sorry that I have to put this on hold again because of the external-ids commit. Please see my comment in https://github.com/eBay/go-ovn/pull/125 for the same commit.

anfredette:
This PR has been reworked on top of PR-125.

PR #126, opened 2021-01-08
better developer experience: ability to run tests inside docker containers

vjayaramrh:
The goal of this PR is to allow for developers (running on non linux systems such as MacBook and who do not have ovs/ovn installed) to be be able to run the tests in containers instead. 
The PR creates a Dockerfile based off the `openvswitch/ovn:2.12_e60f2f2_debian_master` base image and adds go binaries. It also installs `supervisord` to help with running ovn-nb, ovn-sb and ovn-northd in the same container. (Refer http://supervisord.org/introduction.html)

Note: I have an enhancement request in the ovn-org/ovn repo requesting that official docker images for each of the OVN release to quay/docker repo (refer https://github.com/ovn-org/ovn/issues/67)

### DIFF
--- a/utilities/docker/Dockerfile
+++ b/utilities/docker/Dockerfile
@@ -1,0 +1,21 @@
+FROM openvswitch/ovn:2.12_e60f2f2_debian_master
+
+ARG GOVERPKG=go1.13.9.linux-amd64.tar.gz
+RUN apt-get update
+RUN apt-get install -y wget git gcc python3-pip python3-dev build-essential autoconf automake libtool
+
+RUN wget -P /tmp https://dl.google.com/go/$GOVERPKG
+
+RUN tar -C /usr/local -xzf /tmp/$GOVERPKG
+RUN rm /tmp/$GOVERPKG
+
+ENV GOPATH /go
+ENV PATH $GOPATH/bin:/usr/local/go/bin:$PATH
+RUN mkdir -p "$GOPATH/src" "$GOPATH/bin" && chmod -R 777 "$GOPATH"
+
+WORKDIR $GOPATH
+
+RUN apt-get install -y supervisor
+COPY supervisord.conf /etc/supervisor/conf.d/supervisord.conf
+
+ENTRYPOINT ["/usr/bin/supervisord"]
--- a/utilities/docker/supervisord.conf
+++ b/utilities/docker/supervisord.conf
@@ -1,0 +1,11 @@
+[supervisord]
+nodaemon=true
+
+[program:ovn-nb-tcp]
+command=/bin/start-ovn ovn-nb-tcp
+
+[program:ovn-sb-tcp]
+command=/bin/start-ovn ovn-sb-tcp
+
+[program:ovn-northd-tcp]
+command=/bin/start-ovn ovn-northd-tcp
--- a/utilities/docker/README.md
+++ b/utilities/docker/README.md
@@ -1,0 +1,21 @@
+- Goal: Add ability to run the go-ovn tests inside of a docker container which in turn would make it easy to run on 
+  non-linux systems that have docker installed. The docker image installs `go` binaries and sets the `GOPATH` environment
+  variable.
+
+- To build the docker image execute the below on a system where docker is installed
+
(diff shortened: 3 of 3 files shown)


utilities/docker/Dockerfile
```diff
@@ -0,0 +1,21 @@
+FROM openvswitch/ovn:2.12_e60f2f2_debian_master
```
hzhou8, line 1: 2.12 is rather old version, before OVS and OVN split.

utilities/docker/Dockerfile
```diff
@@ -0,0 +1,21 @@
...
+
+ARG GOVERPKG=go1.13.9.linux-amd64.tar.gz
+RUN apt-get update
+RUN apt-get install -y wget git gcc python3-pip python3-dev build-essential autoconf automake libtool
```
hzhou8, line 5: what's the purpose of installing the tools such as gcc, python, build-essential, etc.?

utilities/docker/supervisord.conf
```diff
@@ -0,0 +1,11 @@
...
+nodaemon=true
+
+[program:ovn-nb-tcp]
+command=/bin/start-ovn ovn-nb-tcp
+
+[program:ovn-sb-tcp]
+command=/bin/start-ovn ovn-sb-tcp
+
+[program:ovn-northd-tcp]
+command=/bin/start-ovn ovn-northd-tcp
```
hzhou8, line 11: It seems we don't need northd for go-ovn testing.

hzhou8 (changes requested):
Thanks @vjayaramrh 
It seems the https://github.com/eBay/go-ovn/pull/121 is addressing the problem in a similar way. Could you take a look?
Please also see my comments.

utilities/docker/Dockerfile
vjayaramrh: I used that image as the starting point as it was available in dockerhub at https://hub.docker.com/search?q=openvswitch%2Fovn&type=image, I am hoping that OVN images get regulary pushed as per the request at https://github.com/ovn-org/ovn/issues/67

utilities/docker/Dockerfile
vjayaramrh: > Thanks @vjayaramrh
> It seems the #121 is addressing the problem in a similar way. Could you take a look?
> Please also see my comments.

@hzhou8 appreciate you pointing this out, I have reached out to the PR author to have an initial discussion and look at ways for combining our efforts

vjayaramrh: The test.sh script in the `.travis` folder needs them when executed inside the container

utilities/docker/supervisord.conf
vjayaramrh: correct, but convenient to have a running container built from this image that has all the relevant services running

vjayaramrh:
> Thanks @vjayaramrh
> It seems the #121 is addressing the problem in a similar way. Could you take a look?
> Please also see my comments.

@hzhou8 appreciate you pointing this out, I have reached out to the PR author to have an initial discussion and look at ways for combining our efforts

utilities/docker/Dockerfile
hzhou8: but according to the steps you put in README, the test.sh script is not used. Your approach is supposed to run "go test" directly.

vjayaramrh: The README captures the native `go` way of running tests. Do you think that the README should also capture the alternate way i.e executing the `.travis/test.sh`? If you think that would be useful to capture as well, I can certainly update the README

amorenoz:
@vjayaramrh I think there are some interesting ideas here.
Using a pre-built image would definitely improve the developer experience.

For now, the container does build ovn, only ovs (for the ovsdb-server). The compatibility matrix with ovsdb should be relatively straight forward. But that might change as a result of the discussions in #29. So, when those discussions clear out the go-ovn -> ovn compatibility strategy looking forward, we could improve the current system by using pre-built images. wdyt?

vpickard:
cc @vpickard

PR #130, opened 2021-02-01
add ForceReconnect API

winsopc:
Signed-off-by: Zhen Wang <zhewang@nvidia.com>

### DIFF
--- a/client.go
+++ b/client.go
@@ -227,6 +227,8 @@ type Client interface {
 	// Get PortGroup data structure if it exists
 	PortGroupGet(group string) (*PortGroup, error)
 
+	// Force reconnection to OVN
+	ForceReconnect() error
 	// Close connection to OVN
 	Close() error
 }
@@ -369,6 +371,16 @@ func (c *ovndb) MonitorTables(jsonContext interface{}) (*libovsdb.TableUpdates, 
 			}}
 	}
 	return c.client.Monitor(c.db, jsonContext, requests)
+}
+
+// Disconnect current ovsdb client and reconnct with new one.
+// This API can be used for reconnecting to osvdb with new TLS cert/key pair.
+func (c *ovndb) ForceReconnect() error {
+	// Unregister nofitfier to avoid current client reconnect via callback.
+	err := c.client.Unregister(ovnNotifier{c})
+	c.client.Disconnect()
+	c.reconnect()
+	return err
 }
 
 // TODO return proper error


vtolstov:
LGTM, but can you describe why you need it ?

winsopc:
this API benefits for goovn user ovn-kubernetes to support ovndb TLS certFile and privkeyFile update on purpose.
https://github.com/ovn-org/ovn-kubernetes/blob/master/go-controller/pkg/util/go_ovn.go L87-97.

With new API,  I can use fsnotify watcher to watch the file change and update the tlsconfig.cert and call new API without worry about high layer use goovn.client.

winsopc:
example code to use the api:
`// watchForChanges exits if the configuration file changed.
func updateSslKeyPair(certFile, privKeyFile string, tlsConfig *tls.Config, ovndbclient goovn.Client) error {

	if certFile == "" || privKeyFile == "" {
		return nil
	}
	watcher, err := fsnotify.NewWatcher()
	if err != nil {
		return err
	}

	go func() {
		for {
			select {
			case event, ok := <-watcher.Events:
				if ok && event.Op&fsnotify.Write == fsnotify.Write {
					cert, err := tls.LoadX509KeyPair(certFile, privKeyFile)
					if err != nil {
						klog.Infof("cannot load new cert with cert %s key %s err %s", certFile, privKeyFile, err)
						continue
					}
					if reflect.DeepEqual(tlsConfig.Certificates, []tls.Certificate{cert}) {
						klog.Infof("tls update already finished")
						continue
					}
					tlsConfig.Certificates = []tls.Certificate{cert}
					err = ovndbclient.ForceReconnect()
					klog.Infof("dbclient %p force reconnect with new tlsconfig %p status %v", ovndbclient, tlsConfig, err)
				}
			case err, ok := <-watcher.Errors:
				if ok {
					klog.Errorf("Error watching for changes to ssl key err: %v", privKeyFile, err)
				}
			}
		}
	}()

	if err := watcher.Add(certFile); err != nil {
		return err
	}
	if err := watcher.Add(privKeyFile); err != nil {
		return err
	}
	return nil
}
`

vtolstov:
what you think about adding such example to examples for this function in godoc format?

winsopc:
Sure, I will add it.

amorenoz:
This is a very good feature. Should it be moved to libovsdb?

winsopc:
@amorenoz I think libovsdb needs more code changes to support this compare to support it in go-ovn layer.
The current OvsdbClient struct does not store  *tlsconfig info and endpoints address.

winsopc:
Current solution will trigger connect() to download the DB again.
After studying the go-ovn and libovsdb,  I think current interface can support
the feature without new code changes.
we can rely on underlying rpc2.client reconnect mechanism to reconnect with new cert.

PR #131, opened 2021-02-10
Add support for OVN SB-DB Chassis_Private table

Pardhakeswar:
- ChassisPrivateDel() - Delete row in chassis_private table
    with given chassis name
 - ChassisPrivateList() - List all chassis rows in chassis_private table

### DIFF
--- a/chassis_private.go
+++ b/chassis_private.go
@@ -1,0 +1,141 @@
+/**
+ * Copyright (c) 2020 eBay Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ **/
+
+package goovn
+
+import (
+	"fmt"
+	"github.com/ebay/libovsdb"
+)
+
+// Chassis_Private table OVN SB
+type ChassisPrivate struct {
+	UUID       string
+	ExternalID map[interface{}]interface{}
+	Name       string
+	NbCfg      int
+}
+
+func (odbi *ovndb) chassisPrivateAddImp(chName string,
+	external_ids map[string]string) (*OvnCommand, error) {
+
+	if len(chName) == 0 {
+		return nil, fmt.Errorf("chassis name cannot be empty")
+	}
+
+	namedUUID, err := newRowUUID()
+	if err != nil {
+		return nil, err
+	}
+	// row to insert
+	chPrivate := make(OVNRow)
+	chPrivate["name"] = chName
+
+	if uuid := odbi.getRowUUID(TableChassisPrivate, chPrivate); len(uuid) > 0 {
+		return nil, ErrorExist
+	}
+
+	if external_ids != nil {
+		oMap, err := libovsdb.NewOvsMap(external_ids)
+		if err != nil {
+			return nil, err
+		}
+		chPrivate["external_ids"] = oMap
+	}
+	insertOp := libovsdb.Operation{
+		Op:       opInsert,
+		Table:    TableChassisPrivate,
+		Row:      chPrivate,
+		UUIDName: namedUUID,
+	}
+	operations := []libovsdb.Operation{insertOp}
+	return &OvnCommand{operations, odbi, make([][]map[string]interface{}, len(operations))}, nil
+}
+
+func (odbi *ovndb) chassisPrivateDelImp(name string) (*OvnCommand, error) {
+	var operations []libovsdb.Operation
+
+	condition := libovsdb.NewCondition("name", "==", name)
+	deleteOp := libovsdb.Operation{
+		Op:    opDelete,
+		Table: TableChassisPrivate,
+		Where: []interface{}{condition},
+	}
+	operations = append(operations, deleteOp)
+	return &OvnCommand{operations, odbi, make([][]map[string]interface{}, len(operations))}, nil
+}
+
+func (odbi *ovndb) chassisPrivateListImp() ([]*ChassisPrivate, error) {
+	odbi.cachemutex.RLock()
+	cacheChassisPrivate, ok := odbi.cache[TableChassisPrivate]
+	odbi.cachemutex.RUnlock()
+
+	if !ok {
+		return nil, ErrorSchema
+	}
+
+	listChassisPrivate := make([]*ChassisPrivate, 0, len(cacheChassisPrivate))
+	for uuid := range cacheChassisPrivate {
+		chPrivate, err := odbi.rowToChassisPrivate(uuid)
+		if err != nil {
+			return nil, err
+		}
+		listChassisPrivate = append(listChassisPrivate, chPrivate)
+	}
+	return listChassisPrivate, nil
+}
+
+func (odbi *ovndb) chassisPrivateGetImp(chassis string) ([]*ChassisPrivate, error) {
+	var listChassisPrivate []*ChassisPrivate
+
+	odbi.cachemutex.RLock()
+	cacheChassisPrivate, ok := odbi.cache[TableChassisPrivate]
+	odbi.cachemutex.RUnlock()
+
+	if !ok {
+		return nil, ErrorSchema
+	}
+
+	for uuid, drows := range cacheChassisPrivate {
+		if chName, ok := drows.Fields["name"].(string); ok && chName == chassis {
+			chPrivate, err := odbi.rowToChassisPrivate(uuid)
+			if err != nil {
+				return nil, err
+			}
+			listChassisPrivate = append(listChassisPrivate, chPrivate)
+		}
+	}
+	return listChassisPrivate, nil
+}
+
+func (odbi *ovndb) rowToChassisPrivate(uuid string) (*ChassisPrivate, error) {
+
+	odbi.cachemutex.RLock()
+	cacheChassisPrivate, ok := odbi.cache[TableChassisPrivate][uuid]
+	odbi.cachemutex.RUnlock()
+
+	if !ok {
+		return nil, fmt.Errorf("row in chassis_private with uuid %s not found", uuid)
+	}
+
+	chPrivate := &ChassisPrivate{
+		UUID:       uuid,
+		ExternalID: cacheChassisPrivate.Fields["external_ids"].(libovsdb.OvsMap).GoMap,
+		Name:       cacheChassisPrivate.Fields["name"].(string),
+		NbCfg:      cacheChassisPrivate.Fields["nb_cfg"].(int),
+	}
+	return chPrivate, nil
+}
--- a/chassis_private_test.go
+++ b/chassis_private_test.go
@@ -1,0 +1,74 @@
+/**
+ * Copyright (c) 2020 eBay Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ **/
+
+package goovn
+
+import (
+	"fmt"
+	"testing"
+)
+
+func TestChassisPrivate(t *testing.T) {
+	ovndbapi := getOVNClient(DBSB)
+	t.Logf("Adding row to OVN SB DB chassis_private table")
+	ovn, ok := ovndbapi.(*ovndb)
+	if !ok {
+		t.Fatal(fmt.Errorf("Invalid type assertion"))
+	}
+	ocmd, err := ovn.chassisPrivateAdd(CHASSIS_NAME, nil)
+	if err != nil {
+		t.Fatal(err)
+	}
+	err = ovndbapi.Execute(ocmd)
+	if err != nil {
+		t.Fatalf("err adding row to ovn-sb chassis_private table failed: %v", err)
+	}
+	t.Logf("Adding row to ovn-sb chassis_private table Done")
+
+	t.Logf("Listing all the chassis in ovn-sb chassis_private table")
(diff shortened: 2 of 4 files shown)


client.go
```diff
@@ -375,6 +382,22 @@ func (c *ovndb) MonitorTables(jsonContext interface{}) (*libovsdb.TableUpdates,
...
+// in SB/NB DB schema and currently not in SBTablesOrder or NBTablesOrder
+func (c *ovndb) AddLatestDBSchemaTables() {
+
+	dbSchema, err := c.client.GetSchema(c.db)
+	if err != nil {
+		log.Printf("Failed to get %s db schema: (%v)", c.db, err)
+		return
+	}
+	if c.db == DBSB {
+		if _, ok := dbSchema.Tables[TableChassisPrivate]; ok {
```
girishmg, line 395: I think it is better to use  an array and then walk through the elements of the array, instead of directly checking against TableChassisPrivate.

As in,

var latest SBTables = []string{TableChassisPrivate}

then you can loop through these tables to check whether they are in the dbSchema.Tables.

client.go
```diff
@@ -28,8 +28,9 @@ import (
 )
 
 type EntityType string
-const(
-	PORT_GROUP EntityType = "PORT_GROUP"
+
+const (
+	PORT_GROUP     EntityType = "PORT_GROUP"
 	LOGICAL_SWITCH EntityType = "LOICAL_SWITCH"
```
girishmg, line 34: s/LOICAL/LOGICAL/

client.go
```diff
@@ -330,6 +336,7 @@ func (c *ovndb) reconnect() {
 func (c *ovndb) MonitorTables(jsonContext interface{}) (*libovsdb.TableUpdates, error) {
 	// get the table list based on the DB
 	var tables []string
+	c.AddLatestDBSchemaTables()
```
girishmg, line 339: L339 is inserted in-between L338-344 and the comment @L337 was explaining that. Perhaps you need to move this line to L337

Pardhakeswar: okay, will do the change.

client.go
```diff
@@ -375,6 +384,24 @@ func (c *ovndb) MonitorTables(jsonContext interface{}) (*libovsdb.TableUpdates,
...
+// AddLatestSchemaTables adds new tables that needs to be monitored
+// in SB/NB DB schema and currently not in SBTablesOrder or NBTablesOrder
+func (c *ovndb) AddLatestDBSchemaTables() {
+
+	dbSchema, err := c.client.GetSchema(c.db)
+	if err != nil {
+		log.Printf("Failed to get %s db schema: (%v)", c.db, err)
+		return
+	}
+	if c.db == DBSB {
```
vtolstov, line 396: return early
if c.db != DBSB {
  return
}

chassis_private.go
```diff
@@ -0,0 +1,140 @@
...
+		}
+	}
+	return listChassisPrivate, nil
+}
+
+func (odbi *ovndb) rowToChassisPrivate(uuid string) (*ChassisPrivate, error) {
+
+	cacheChassisPrivate, ok := odbi.cache[TableChassisPrivate][uuid]
+	if !ok {
+		return nil, fmt.Errorf("Row in Chassis_Private with uuid%s not found", uuid)
```
vtolstov, line 130: uuid%s ?

vtolstov: also use lowercase error messages

vtolstov (changes requested):

Pardhakeswar:
Thanks @vtolstov for the review. Updated the PR with requested changes.

chassis_private.go
```diff
@@ -0,0 +1,140 @@
...
+	return listChassisPrivate, nil
+}
+
+func (odbi *ovndb) chassisPrivateGetImp(chassis string) ([]*ChassisPrivate, error) {
+	var listChassisPrivate []*ChassisPrivate
+
+	odbi.cachemutex.RLock()
+	defer odbi.cachemutex.RUnlock()
+
+	cacheChassisPrivate, ok := odbi.cache[TableChassisPrivate]
```
vtolstov, line 108: will be better instead of defer and cacheChassisPrivate, ok := odbi.cache[TableChassisPrivate]
use something like this
odbi.cachemutex.RLock()
cacheChassisPrivate, ok := odbi.cache[TableChassisPrivate]
odbi.cachemutex.RUnlock()
if !ok {
....
}

in small functions better to avoid defer usage

chassis_private.go
```diff
@@ -0,0 +1,140 @@
...
+			}
+			listChassisPrivate = append(listChassisPrivate, chPrivate)
+		}
+	}
+	return listChassisPrivate, nil
+}
+
+func (odbi *ovndb) rowToChassisPrivate(uuid string) (*ChassisPrivate, error) {
+
+	cacheChassisPrivate, ok := odbi.cache[TableChassisPrivate][uuid]
```
vtolstov, line 128: odbi.cachemutex.RLock()
...
odbi.cachemutex.RUnlock()

vtolstov (changes requested):
needs to do small changes and will be ready to merge

chassis_private.go
Pardhakeswar: @vtolstov **rowToChassisPrivate** is being called from **ChassisPrivateGet** & **ChassisPrivateGet**
adding  odbi.cachemutex.RLock in **rowToChassisPrivate** will lead to scenario right as we already have locked **rowToChassisPrivate**?

Pardhakeswar:
@vtolstov updated the PR.

vtolstov (approved):

vtolstov:
LGTM, @noah8713 do you want to review?

chassis_private_test.go
```diff
@@ -0,0 +1,69 @@
...
+	if len(chassisPrivate) != 1 {
+		t.Fatalf("err getting chassis in ovn-sb chassis_private table, total:%v",
+			len(chassisPrivate))
+	}
+	chName := chassisPrivate[0].Name
+	t.Logf("Chassis found in chassis-private table:%+v", chName)
+
+	// deleting the row in chassis_private table
+	t.Logf("Deleting row in OVN SB DB chassis_private table:%v", chName)
+	ocmd, err = ovndbapi.ChassisDel(chName)
```
noah8713, line 61: it should be chassisprivatedel not chassisdel.

client.go
```diff
@@ -28,9 +28,10 @@ import (
 )
 
 type EntityType string
-const(
-	PORT_GROUP EntityType = "PORT_GROUP"
-	LOGICAL_SWITCH EntityType = "LOICAL_SWITCH"
+
```
noah8713, line 31: fmt should be part of different commit.

client.go
```diff
@@ -375,6 +384,25 @@ func (c *ovndb) MonitorTables(jsonContext interface{}) (*libovsdb.TableUpdates,
 	return c.client.Monitor(c.db, jsonContext, requests)
 }
 
+// AddLatestSchemaTables adds new tables that needs to be monitored
```
noah8713, line 387: why need AddLatestDBSchemaTables ?
Adding new table should be part of NB/SB table order as we keep adding tables on need basis. 

Client can watch on any tables on need basis

client.go
```diff
@@ -747,3 +787,7 @@ func (c *ovndb) sbGlobalAdd(options map[string]string) (*OvnCommand, error) {
 func (c *ovndb) sbGlobalDel() (*OvnCommand, error) {
 	return c.sbGlobalDelImp()
 }
+
+func (c *ovndb) chassisPrivateAdd(name string, external_ids map[string]string) (*OvnCommand, error) {
```
noah8713, line 791: Please move chassisPrivateAdd naear L433 to avoid scattering api calls in different places in function

common.go
```diff
@@ -81,3 +82,7 @@ var SBTablesOrder = []string{
 	TableEncap,
 	TableSBGlobal,
 }
+
+var LatestSBTablesOrder = []string{
```
noah8713, line 86: What is the intent for LatestSBTablesOrder ?TableChassisPrivate should be in SBTablesOrder

client.go
girishmg: @noah8713 the Chassis_Private Tables is not present in all releases of OVN. It is only present in December-2020 release. So, we add that table if OVN SB driver returns that table in its schema. Hence, it needs to be dynamic.

common.go
girishmg: The intent again is that ChassisPrivate table was only recently added. It doesn't exist on the old releases of OVN.

client.go
Pardhakeswar: @noah8713 **chassisPrivateAdd** api is only used as helper function in unit-tests so that was the reason I added in the section of api's which are used as helpers for unit tests.

common.go
noah8713: thanks @girishmg for details. So if new table gets added to schema everytime, we keep changing go-ovn with latest* table? I think thats not the end goal. Also our current fork runs against latest master so should be fine. Please rectify if I got it something the other way,

client.go
noah8713: Thanks. However, any api exposed in client should follow CRUD order. Also please check as CI is failing too:
https://travis-ci.org/github/eBay/go-ovn/builds/758743285?utm_source=github_status&utm_medium=notification

```
./chassis_private_test.go:26:23: ovndbapi.chassisPrivateAdd undefined (type Client has no field or method chassisPrivateAdd)
note: module requires Go 1.12FAIL	github.com/ebay/go-ovn [build failed]
```

client.go
noah8713: Seems then such optimization needs to be for any new table coming up. I think we add table api support for any new table being exposed in OVN. We can add a check in CRUD calls if underlying ovn doesnt have such table exposed. Also client has the flexibility to watch on tables based on their underlying ovn version. So little hesitating as to why go-ovn should worry about if table is missing in actual ovn that its connecting to.

noah8713:
@Pardhakeswar : Thanks for your PR. Please resolve pending/closed comments that are addressed/raised by @girishmg @vtolstov  as looking at the PR shows 10s of comments all open. Would be easy to address the new pending accordingly.

client.go
Pardhakeswar: @noah8713 these fmt changes aren't needed with this commit?

client.go
girishmg: @noah8713 we will break existing clients connecting to old server if we blindly update the {SB|NB}TablesOrder. Other way to solve would be to always update the  {SB|NB}TablesOrder array and remove the tables that the remote OVN server doesn't support just before we call MonitorTables().

client.go
noah8713: no

client.go
noah8713: This still shows failure in CI. Did you push the fix?

noah8713: Please move this helper function with CRUD calls in line

Pardhakeswar: Not yet, waiting for the resolution of the discussion on LatestSBTablesOrder

client.go
noah8713: @girishmg : Thanks. I think clients can still pin go-ovn to particular commit to support prior/onwards dec 2020 remote OVN releases. I second for the approach of doing check if new table actually exists before calling monitoring tables. I am ok if that can be incremental PR for client.go to ensure that safety check for any new future table addition/deletion in OVN and add chassisprivate to {SB|NB}TablesOrder.

Pardhakeswar:
@noah8713 addressed the comments and updated the PR accordingly.

client.go
hzhou8: Hi @girishmg @noah8713 thanks for the good discussion. I agree with @noah8713 that AddLatestDBSchemaTables() and LatestSBTablesOrder look weird. In fact, I think not only LatestSBTablesOrder, but NBTablesOrder/SBTablesOrder are not needed. They are used only in MonitorTables() for validating if a table is valid. In my opinion the available tables don't need to be hardcoded. I'd suggest removing NBTablesOrder/SBTablesOrder all together, but it should be done in a separate commit.

chassis_private_test.go
```diff
@@ -0,0 +1,74 @@
...
+	t.Logf("Listing all the chassis in ovn-sb chassis_private table")
+	chassisPrivate, err := ovndbapi.ChassisPrivateList()
+	if err != nil {
+		t.Fatal(err)
+	}
+	if len(chassisPrivate) != 1 {
+		t.Fatalf("err listing chassis in ovn-sb chassis_private table, total:%v",
+			len(chassisPrivate))
+	}
+	t.Logf("Single chassis found in ovn-sb chassis_private table: %+v", chassisPrivate)
```
noah8713, line 50: chassisPrivate is a list. So use chassisPrivate[0].Name

common.go
noah8713, line 86: remove this section.

client.go
noah8713: Sounds good. Thanks @hzhou8 . @Pardhakeswar : Please remove the add/latestdb part and just add chassisprivate as a regular table support following other tables source as a ref. @girishmg  and team please feel free to file an incremental PR to remove NB/SBTables in general as an improvement for dynamic table realization.

noah8713 (approved):
lgtm

client.go
girishmg: @hzhou8 as I understand go-ovn does not support all the tables in SB/NB since there are no APIs to do CRUD operations against the unsupported tables. 

For example:  If a client using go-ovn requests for OVN SB's Datapath_binding, then go-ovn will request for that table to be monitored and it will get updates from the SB server, however they can't do much with those updates since there are no APIs.

My understanding was that those tables clearly said what tables are supported for CRUD operation by this library.

hzhou8: @girishmg I think the APIs should clearly tell what's supported, and when a user specifically request for monitoring a specific table, they must have clear expectation. It doesn't make much sense for a user to request a table that is not even supported by the API. Or maybe your real concern is the default behavior, i.e. when user doesn't specify which tables to monitor, falling back to monitor-all would incur some more waste?

girishmg: @hzhou8 correct the memory footprint of the clients would explode, for example, with Port_Binding and Datapth_Binding tables from OVN SB DB in a large scale cluster.

By default,  this library should monitor the tables for which there is support for CRUD operations and the end users can filter the table further from that supported list. I think the {SB|NB}TablesOrder enabled it.

Now, with ChassisPrivate table added to SBTablesOrder array and if someone upgrades to latest go-ovn library to get the recently added PortGroup and ACL APIs, then they can't use this library against their old release of OVN since it doesn't support ChassisPrivate table.

The initial PR, added ChassisPrivate table only if the SB Schema had support for it.

hzhou8: @girishmg Thanks for clarifying and now I understand your concern. So I think it makes sense to keep the NB/SBTablesOrder arrays for the sake of saving cost when users don't explicitly specify what tables are interested. While for the backward compatibility, it is more straightforward to just keep the tables in NB/SBTablesOrder matching the APIs we support and as you proposed: "remove the tables that the remote OVN server doesn't support just before we call MonitorTables()" sounds good to me.

girishmg: @hzhou8 great! @Pardhakeswar can you please take care of this? @noah8713 FYI

PR #132, opened 2021-02-16
add universal handlers for LS external_ids and other_config records 

nvity:
in some cases, we need to update other_config record with key/values specific to a logical entity. For example, DHCP parameters such as subnet CIDR and exclude_ips live in logical switch'es other_config record. Currently, there's no API methods to update other_config. This PR provides "Add" and "Del" methods (LSAux{Add,Del}) that can operate either on external_ids or other_config records by providing the corresponding parameter. Also, added LSP Type setter, so we can modify port type (e.g, when we want to set localnet type for the port)

### DIFF
--- a/client.go
+++ b/client.go
@@ -64,6 +64,8 @@ type Client interface {
 	LSPDel(lsp string) (*OvnCommand, error)
 	// Set addressset per lport
 	LSPSetAddress(lsp string, addresses ...string) (*OvnCommand, error)
+	// Set lport type
+	LSPSetType(lsp string, portType string) (*OvnCommand, error)
 	// Set port security per lport
 	LSPSetPortSecurity(lsp string, security ...string) (*OvnCommand, error)
 	// Get all lport by lswitch
@@ -453,6 +455,10 @@ func (c *ovndb) LSPDel(lsp string) (*OvnCommand, error) {
 
 func (c *ovndb) LSPSetAddress(lsp string, addresses ...string) (*OvnCommand, error) {
 	return c.lspSetAddressImp(lsp, addresses...)
+}
+
+func (c *ovndb) LSPSetType(lsp string, portType string) (*OvnCommand, error) {
+	return c.lspSetTypeImp(lsp, portType)
 }
 
 func (c *ovndb) LSPSetPortSecurity(lsp string, security ...string) (*OvnCommand, error) {
--- a/client.go
+++ b/client.go
@@ -28,8 +28,9 @@ import (
 )
 
 type EntityType string
-const(
-	PORT_GROUP EntityType = "PORT_GROUP"
+
+const (
+	PORT_GROUP     EntityType = "PORT_GROUP"
 	LOGICAL_SWITCH EntityType = "LOICAL_SWITCH"
 )
 
@@ -48,6 +49,10 @@ type Client interface {
 	LSExtIdsAdd(ls string, external_ids map[string]string) (*OvnCommand, error)
 	// Del external_ids from logical_switch
 	LSExtIdsDel(ls string, external_ids map[string]string) (*OvnCommand, error)
+	// Add external_ids or other_config records to logical switch
+	LSAuxAdd(ls string, auxConf map[string]string, auxTable string) (*OvnCommand, error)
+	// Del external_ids or other_config from logical_switch
+	LSAuxDel(ls string, auxConf map[string]string, auxTable string) (*OvnCommand, error)
 	// Link logical switch to router
 	LinkSwitchToRouter(lsw, lsp, lr, lrp, lrpMac string, networks []string, externalIds map[string]string) (*OvnCommand, error)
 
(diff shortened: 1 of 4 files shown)


vgozer:
any chance to review (@hzhou8 @vtolstov ...)? can resubmit to avoid merge conflict (popped up after this PR was submitted)

client.go
```diff
@@ -28,8 +28,9 @@ import (
 )
 
 type EntityType string
-const(
-	PORT_GROUP EntityType = "PORT_GROUP"
+
+const (
+	PORT_GROUP     EntityType = "PORT_GROUP"
 	LOGICAL_SWITCH EntityType = "LOICAL_SWITCH"
```
girishmg, line 34: needs rebase, since the above got fixed with this commit https://github.com/eBay/go-ovn/commit/16f56680e2c066dc3394037587ca9bf24e04f646

logical_switch.go
```diff
@@ -331,6 +331,70 @@ func (odbi *ovndb) lsExtIdsDelImp(ls string, external_ids map[string]string) (*O
...
+		return nil, fmt.Errorf("%s is neither other_config or external_ids", auxTable)
+	}
+
+	var operations []libovsdb.Operation
+	row := make(OVNRow)
+	row["name"] = ls
+	lsuuid := odbi.getRowUUID(TableLogicalSwitch, row)
+	if len(lsuuid) == 0 {
+		return nil, ErrorNotFound
+	}
```
girishmg, line 346: Is L343-346, really required? Say, that logical switch is bogus, eventually the commaunt would fail when we execute it later, isn't it? 

Also, this check might succeed now, but little later when the command is run it could be that the logical switch has been deleted.

logical_switch.go
```diff
@@ -331,6 +331,70 @@ func (odbi *ovndb) lsExtIdsDelImp(ls string, external_ids map[string]string) (*O
...
+	if len(auxConf) == 0 {
+		return nil, fmt.Errorf("%s map is nil or empty", auxTable)
+	}
+	mutateSet, err := libovsdb.NewOvsMap(auxConf)
+	if err != nil {
+		return nil, err
+	}
+	mutation := libovsdb.NewMutation(auxTable, opInsert, mutateSet)
```
girishmg, line 354: from RFC 7047, we have:

   <mutation>
      A 3-element JSON array of the form [<column>, <mutator>, <value>]
      
      map
         <mutator> must be "insert" or "delete".

         If <mutator> is "insert", then each of the key-value pairs in
         the map in <value> is added to <column> only if its key is not
         already present.  The required type of <value> is slightly
         relaxed, in that it may have fewer than the minimum number of
         elements specified by the column's type.

Say, if the external_ids column already has that key, then the call to this API will not update that key from the definition above.

So, should we first get the current external_ids, merge the to-be-set external_ids and then proceed? Am I missing something?

logical_switch.go
```diff
@@ -331,6 +331,70 @@ func (odbi *ovndb) lsExtIdsDelImp(ls string, external_ids map[string]string) (*O
...
+		Op:        opMutate,
+		Table:     TableLogicalSwitch,
+		Mutations: []interface{}{mutation},
+		Where:     []interface{}{condition},
+	}
+	operations = append(operations, mutateOp)
+	return &OvnCommand{operations, odbi, make([][]map[string]interface{}, len(operations))}, nil
+}
+
+// lsAuxDelImp is for removing records from external_ids or other_config tables
```
girishmg, line 366: s/removing key values from external...  columns/?

logical_switch.go
```diff
@@ -331,6 +331,70 @@ func (odbi *ovndb) lsExtIdsDelImp(ls string, external_ids map[string]string) (*O
...
+func (odbi *ovndb) lsAuxDelImp(ls string, auxConf map[string]string, auxTable string) (*OvnCommand, error) {
+	if (auxTable != "external_ids") && (auxTable != "other_config") {
+		return nil, fmt.Errorf("%s is neither other_config or external_ids", auxTable)
```
girishmg, line 369: rename auxTable, same as before?

logical_switch.go
```diff
@@ -331,6 +331,70 @@ func (odbi *ovndb) lsExtIdsDelImp(ls string, external_ids map[string]string) (*O
...
+	}
+
+	var operations []libovsdb.Operation
+	row := make(OVNRow)
+	row["name"] = ls
+	lsuuid := odbi.getRowUUID(TableLogicalSwitch, row)
+	if len(lsuuid) == 0 {
+		return nil, ErrorNotFound
+	}
```
girishmg, line 378: L375-378: same comment as before.

girishmg (changes requested):

logical_switch.go
vgozer: if key is already present, then there's no other way of updating its value but to remove it and then insert this key back with new value, right?

logical_switch.go
```diff
@@ -331,6 +331,70 @@ func (odbi *ovndb) lsExtIdsDelImp(ls string, external_ids map[string]string) (*O
 	return &OvnCommand{operations, odbi, make([][]map[string]interface{}, len(operations))}, nil
 }
 
+// lsAuxAddImp is for adding records to external_ids or other_config tables
+func (odbi *ovndb) lsAuxAddImp(ls string, auxConf map[string]string, auxTable string) (*OvnCommand, error) {
```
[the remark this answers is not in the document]
vgozer: ack

logical_switch.go
```diff
@@ -331,6 +331,70 @@ func (odbi *ovndb) lsExtIdsDelImp(ls string, external_ids map[string]string) (*O
...
+	if (auxTable != "external_ids") && (auxTable != "other_config") {
+		return nil, fmt.Errorf("%s is neither other_config or external_ids", auxTable)
```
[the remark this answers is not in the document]
vgozer: ack

client.go
vgozer: will close this PR and resubmit with fixes (this time from a topic branch)

logical_switch.go
vgozer: makes sense. will remove

logical_switch.go
vgozer: ack

logical_switch.go
vgozer: ack

logical_switch.go
vgozer: ack

PR #133, opened 2021-02-18
Check tables in SBTablesOrder & NBTablesOrder with current ovn-db schema

Pardhakeswar:

### DIFF
--- a/client.go
+++ b/client.go
@@ -338,15 +338,30 @@ func (c *ovndb) reconnect() {
 	}()
 }
 
-func (c *ovndb) MonitorTables(jsonContext interface{}) (*libovsdb.TableUpdates, error) {
-	// get the table list based on the DB
+// filterTablesFromSchema checks whether tables in
+// NBTablesOrder / SBTablesOrder exists in current ovn-db schema
+func (c *ovndb) filterTablesFromSchema() []string {
 	var tables []string
+
+	// get the table list based on the DB
 	if c.db == DBNB {
 		tables = NBTablesOrder
 	} else {
 		tables = SBTablesOrder
 	}
 
+	dbSchema := c.GetSchema()
+	schemaTables := make([]string, 0)
+	for _, table := range tables {
+		if _, ok := dbSchema.Tables[table]; ok {
+			schemaTables = append(schemaTables, table)
+		}
+	}
+	return schemaTables
+}
+
+func (c *ovndb) MonitorTables(jsonContext interface{}) (*libovsdb.TableUpdates, error) {
+	tables := c.filterTablesFromSchema()
 	// verify whether user specified table and its columns are legit
 	if len(c.tableCols) != 0 {
 		supportedTableMaps := make(map[string]bool)


Pardhakeswar:
@girishmg @noah8713 can you PTAL?

client.go
```diff
@@ -335,15 +335,34 @@ func (c *ovndb) reconnect() {
...
 	var tables []string
+
+	// get the table list based on the DB
 	if c.db == DBNB {
 		tables = NBTablesOrder
 	} else {
 		tables = SBTablesOrder
 	}
 
+	dbSchema, err := libovsdb.GetSchema(c.db)
```
noah8713, line 351: Did you test this in your local as CI is failing?
```
./client.go:351:19: undefined: libovsdb.GetSchema
note: module requires Go 1.12FAIL	github.com/ebay/go-ovn [build failed]
```

Seems we need to handle such error in CI too to avoid showing exited with 0. instead of failure.

Pardhakeswar: @noah8713 updated it.

client.go
```diff
@@ -335,15 +335,34 @@ func (c *ovndb) reconnect() {
 	}()
 }
 
-func (c *ovndb) MonitorTables(jsonContext interface{}) (*libovsdb.TableUpdates, error) {
-	// get the table list based on the DB
+// checkSchemaTables checks whether tables in
+// NBTablesOrder / SBTablesOrder exists in current ovn-db schema
+func (c *ovndb) checkSchemaTables() []string {
```
girishmg, line 340: @Pardhakeswar I would rename this to be

filterTablesFromSchema()

client.go
```diff
@@ -335,15 +335,34 @@ func (c *ovndb) reconnect() {
...
+	schemaTables := make([]string, 0)
```
girishmg, line 341: move L341 close to where it is being used

client.go
```diff
@@ -335,15 +335,34 @@ func (c *ovndb) reconnect() {
...
 	var tables []string
+
+	// get the table list based on the DB
 	if c.db == DBNB {
 		tables = NBTablesOrder
 	} else {
 		tables = SBTablesOrder
 	}
 
+	dbSchema, err := c.client.GetSchema(c.db)
```
girishmg, line 351: why are you calling c.client.GetSchema(), again? 
c.client.Schema[c.db] should already have that, isn't it?

girishmg: you can then remove lines 352-355

girishmg (changes requested):

client.go
Pardhakeswar: Missed that, use that.

hzhou8 (approved):
LGTM

client.go
```diff
@@ -335,15 +335,30 @@ func (c *ovndb) reconnect() {
...
 	var tables []string
+
+	// get the table list based on the DB
 	if c.db == DBNB {
 		tables = NBTablesOrder
 	} else {
 		tables = SBTablesOrder
 	}
 
+	dbSchema := c.client.Schema[c.db]
```
noah8713, line 350: Just small change. Use GetSchema() as support it already added. Rest LGTM

Pardhakeswar: @noah8713 updated it.

noah8713: thanks. lgtm

PR #134, opened 2021-02-20
Add GetSchema() API

Pardhakeswar:

### DIFF
--- a/client.go
+++ b/client.go
@@ -241,6 +241,9 @@ type Client interface {
 
 	// Close connection to OVN
 	Close() error
+
+	// GetSchema() returns ovn-db schema
+	GetSchema() libovsdb.DatabaseSchema
 }
 
 var _ Client = &ovndb{}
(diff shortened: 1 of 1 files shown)


client.go
```diff
@@ -389,6 +392,10 @@ func (c *ovndb) Close() error {
 	return nil
 }
 
+func (c *ovndb) GetSchema() (libovsdb.DatabaseSchema, error) {
+        return c.client.Schema[c.db], nil
```
girishmg, line 396: @Pardhakeswar  I don't think you need to return error at all.. since it is going to be `nil` all the time

Pardhakeswar: Updated it @girishmg

girishmg:
/lgtm

@hzhou8 @noah8713 @vtolstov the background behind this is that in `ovn-k8s` we want to make sure that the OVN backend supports a table before we call APIs. Having this API gives the client a chance to make such decisions.

noah8713:
Thanks @Pardhakeswar @girishmg  . Looks neat. /lgtm.
Seems https://github.com/eBay/go-ovn/pull/133 can re-use this as it validates current table with getschema.

Will wait a bit if @hzhou8 @vtolstov have any other suggestions. Can be merged accordingly.

PR #135, opened 2021-02-20
Add support for LS other_config; setter for LSP type

vgozer:
1) New methods for Add/Del key/val to LS external_ids, other_config. The latter is needed to set DHCP parameters.
2) New method to set LSP type (e.g., needed for 'localnet')

### DIFF
--- a/client.go
+++ b/client.go
@@ -49,6 +49,10 @@ type Client interface {
 	LSExtIdsAdd(ls string, external_ids map[string]string) (*OvnCommand, error)
 	// Del external_ids from logical_switch
 	LSExtIdsDel(ls string, external_ids map[string]string) (*OvnCommand, error)
+	// Add keys/vals to logical switch external_ids or other_config records
+	LSAuxAdd(ls string, auxConf map[string]string, auxColumn string) (*OvnCommand, error)
+	// Remove keys/vals from logical switch external_ids or other_config records
+	LSAuxDel(ls string, auxConf map[string]string, auxColumn string) (*OvnCommand, error)
 	// Link logical switch to router
 	LinkSwitchToRouter(lsw, lsp, lr, lrp, lrpMac string, networks []string, externalIds map[string]string) (*OvnCommand, error)
 
@@ -444,6 +448,14 @@ func (c *ovndb) LSExtIdsAdd(ls string, external_ids map[string]string) (*OvnComm
 
 func (c *ovndb) LSExtIdsDel(ls string, external_ids map[string]string) (*OvnCommand, error) {
 	return c.lsExtIdsDelImp(ls, external_ids)
+}
+
+func (c *ovndb) LSAuxAdd(ls string, auxConf map[string]string, auxColumn string) (*OvnCommand, error) {
+	return c.lsAuxAddImp(ls, auxConf, auxColumn)
+}
+
+func (c *ovndb) LSAuxDel(ls string, auxConf map[string]string, auxColumn string) (*OvnCommand, error) {
+	return c.lsAuxDelImp(ls, auxConf, auxColumn)
 }
 
 func (c *ovndb) LSPGet(lsp string) (*LogicalSwitchPort, error) {
--- a/client.go
+++ b/client.go
@@ -66,6 +66,8 @@ type Client interface {
 	LSPSetAddress(lsp string, addresses ...string) (*OvnCommand, error)
 	// Set port security per lport
 	LSPSetPortSecurity(lsp string, security ...string) (*OvnCommand, error)
+	// Set lport type
+	LSPSetType(lsp string, portType string) (*OvnCommand, error)
 	// Get all lport by lswitch
 	LSPList(ls string) ([]*LogicalSwitchPort, error)
 
@@ -480,6 +482,10 @@ func (c *ovndb) LSPSetAddress(lsp string, addresses ...string) (*OvnCommand, err
 
 func (c *ovndb) LSPSetPortSecurity(lsp string, security ...string) (*OvnCommand, error) {
 	return c.lspSetPortSecurityImp(lsp, security...)
+}
+
+func (c *ovndb) LSPSetType(lsp string, portType string) (*OvnCommand, error) {
+	return c.lspSetTypeImp(lsp, portType)
 }
 
 func (c *ovndb) LSPSetDHCPv4Options(lsp string, options string) (*OvnCommand, error) {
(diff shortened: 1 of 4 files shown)


logical_switch.go
```diff
@@ -331,6 +331,66 @@ func (odbi *ovndb) lsExtIdsDelImp(ls string, external_ids map[string]string) (*O
...
+	mutateSet, err := libovsdb.NewOvsMap(auxConf)
+	if err != nil {
+		return nil, err
+	}
+
+	var operations []libovsdb.Operation
+	row := make(OVNRow)
+	row["name"] = ls
+
+	mutation := libovsdb.NewMutation(auxCol, opInsert, mutateSet)
```
vgozer, line 352: per previous discussion with @girishmg, our understanding (per RFC 7047) is that if the key already exists, its value will not be updated. a special handling of this situation will be required - do we read back the entire record, scan all the keys, replace old values with the new ones, add new K/Vs, insert the merged record? Or let user deal with it. Otherwise, the new values for the old keys will be silently ignored, if the above assumption is correct.

amorenoz: If this is something related to RFC7047 (which I agree it is), should we add this functionality to libovsdb to handle it? That way, all present and future ovn (and non-ovn) objects can benefit from it. WDYT?

vgozer: +1 for libovsdb handling it. makes more sense than on the client side, IMHO.

amorenoz: I'm working on an object-based API for libovsdb and I was considering maybe it would be a good idea to implement operation logic, so we could end up with something like this: https://gorm.io/docs/update.html#Update-single-column

Even before that, we could add it to https://github.com/eBay/libovsdb/pull/30

hzhou8: If we want to overwrite existing keys, according to RFC7047, we will have to use "update" instead of "mutate". We don't have to "read back the entire record" because it is already in the local cache. We can't do this in libovsdb right now because libovsdb doesn't maintain any cache. However, I agree that the object-based API may take this into consideration, but I haven't reviewed it, so not really sure.

vgozer: so, if the consensus is that go-ovn *Imp() are not the best place to deal with this right now, let's leave them as-is and merge this PR?

vgozer: @hzhou8 @vtolstov @noah8713 bump ... any chance for a merge?

amorenoz: @vgozer, I think @hzhou8 is right, we can leverage the cache in go-ovn to figure out if we need to 'mutate' or 'update'. In order to implement the desired functionality you really need to do that or else it would silently ignore existing keys.

client.go
```diff
@@ -49,6 +49,10 @@ type Client interface {
 	LSExtIdsAdd(ls string, external_ids map[string]string) (*OvnCommand, error)
 	// Del external_ids from logical_switch
 	LSExtIdsDel(ls string, external_ids map[string]string) (*OvnCommand, error)
+	// Add keys/vals to logical switch external_ids or other_config records
+	LSAuxAdd(ls string, auxConf map[string]string, auxColumn string) (*OvnCommand, error)
```
hzhou8, line 53: It is embarrassing that we didn't have a consistent way to manipulate fields like external_ids and other_config. Now that we already had LSExtIdsAdd/LSExtIdsDel for LS external_ids, I think it may be better to avoid adding overlapping APIs just for LS.

There are two options:

Option1: We can just add LSOtherConfigAdd (in fact it is better to be named as xxxSet, which should overwrite existing key/value) and LSOtherConfigDel. (we need to fix the behavior of LSExtIdsAdd so that it takes care of existed key updating)

Option2: It may be even better to just add common APIs for these common columns to support all tables instead of separate APIs for each table. These APIs can just take "tableName" as an additional parameter, such as:

* AuxKeyValSet(table_name string, row_name string, auxColumn string, auxConf map[string]string) (*OvnCommand, error)
* AuxKeyValDel(table_name string, row_name string, auxColumn string, auxConf map[string]string) (*OvnCommand, error)

What do you think?

hzhou8 (changes requested):

logical_switch.go
vgozer: @hzhou8 agree, let's decide on the proper way - this PR was meant to be a quick fix, but I hit a bunch of other issues, had to work them around with ... ovn-nbctl exec from the go code (sigh). Now we can spend more time on this and DTRT :)

logical_switch.go
```diff
@@ -331,6 +331,66 @@ func (odbi *ovndb) lsExtIdsDelImp(ls string, external_ids map[string]string) (*O
...
+// lsAuxDelImp is for removing keys/vals from either external_ids or other_config records
+func (odbi *ovndb) lsAuxDelImp(ls string, auxConf map[string]string, auxCol string) (*OvnCommand, error) {
+	if (auxCol != "external_ids") && (auxCol != "other_config") {
+		return nil, fmt.Errorf("%s is neither other_config nor external_ids", auxCol)
+	}
+	if len(auxConf) == 0 {
+		return nil, fmt.Errorf("%s map is nil or empty", auxCol)
+	}
+
+	mutateSet, err := libovsdb.NewOvsMap(auxConf)
```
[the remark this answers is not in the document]
vgozer: @hzhou8 that'd be useful, IMHO - empty map means "delete" all k/v for update

client.go
vgozer: @hzhou8 IMHO option2 is the best, to solve this once and for all (tables). Maybe even a 3rd method - AuxKeyValGet()?

logical_switch.go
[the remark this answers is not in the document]
hzhou8: > empty map means "delete" all k/v

Not sure if this a good idea. The API name suggests "delete" the stuff in the map, so users may be easily think empty means delete nothing.

> key: "" means delete this k/v

Yes, this is what I proposed, but I thought about a corner case when we do allow empty string "" as a value, and now we only wants to delete the key iff the value is exactly "". Unfortunately there is no "nil" for string in Go. Another option is to use map[string]*string as the parameter type, but it would seem inconvenience for the normal use cases. Any other ideas?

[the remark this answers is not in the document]
vgozer: yeah, sending a nil/empty map (especially, nil) and removing all K/Vs would be too aggressive/dangerous. WRT empty string vs nil ptr - the latter seems ok to me. other options are more complex/less attractive (if we ignore the corner case with "" as value). nil ptr for string is the simplest, if the delete key semantics is useful, I'd go for it.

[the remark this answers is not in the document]
hzhou8: ok, so we may implement the generic k/v del API using map[string]*string.

vgozer:
closing this PR, submitting a new one that does what has been proposed in the review comments.

vgozer:
closing this PR, submitting a new one with the proposed changes

PR #136, opened 2021-02-25
(WIP) RFC: Add ORM API

amorenoz:
This PR is based on https://github.com/eBay/libovsdb/pull/31

## Why add a ORM API? ##

- It makes it much easier to interact with go-ovn as clients do not have to deal with libovsdb specific objects (such as `OvsMap` and `OvsSet`.
- It automatically enables generic schema-based parsing of any object in any table in any database without needing to write yet another `RowTo{X}` function.
- It can enable considerable reduction of LOC, while having a the same (or even more) functionality
- It enables the automatic generation of native bindings so clients can keep track of changes in their databases
- It also enables clients to specify what fields they want to read based on their provided models, potentially saving some cycle

## How it works? ##
This work is heavily based on https://github.com/eBay/libovsdb/pull/31 and https://github.com/eBay/libovsdb/pull/30.

We define a simple Model interface
```
type TableName string
type Model interface {
	Table() TableName
}
```
Any struct that implements this interface and tags its fields [accordingly](https://github.com/eBay/libovsdb/pull/31), can be used to read from / write to a table in the database.

A full `DBModel` is also a simple struct
```
// DBModel is a Database model
type DBModel struct {
	types map[TableName]reflect.Type
}
```
For each table, it has a pointer to it's ORM types

By passing a DBModel to the go-ovn client, it could be able to:

- Parse all incoming rows, transform them into instances of the right struct and store them that way in the cache
- Pass a pointer to such struct on the callbacks that now can be generic:
```
type OVNORMSignal interface {
	OnCreated(Model)
	OnDeleted(Model)
}
```
- The client can just cast the incoming Model to the right struct and use it (see example in PR)
- Implement a generic Add, Update, Delete functions (no need to add per-table logic any more)

## Examples ##
The best example can be seen [here](https://github.com/amorenoz/ovnmodel), more specifically in this 70 lines of code: https://github.com/amorenoz/ovnmodel/blob/main/cmd/ovnmon/ovnmon.go
That small program is able to receive native objects for the entire nbctl database without parsing data
Also, a code generator that genereates valid `Model`s and `DBModel` from a given ovsschema can be seen [here](https://github.com/amorenoz/ovnmodel/tree/main/cmd/modelgen)

## Status ##
This PR is WIP.
Only reading from the database (i.e: generating events as a result of cache-population) is supported (see examples).
It implements a generic List() command.
My intention is to add Create() Delete() and Update() functions to it but sending it out early on to get feedback


amorenoz:
cc @aojea

hzhou8 (commented):
Thanks @amorenoz for the great work. Could you add more in the example to demonstrate how all CRUD operations would look like?

dave-tucker:
@amorenoz is this up-to-date? I can't see the `Create()` and `Get()` implementations you linked in #145.

I'm also curious on how you plan on handling mutations. A very common use case would be as follows (in pseudo-go)
```go

client := goovn.Client()
// i need to create a logical router static route and reference it from a logical router
// i know the logical router name though

var logicalRouter goovn.LogicalRouter
// I'm not sure what the Get syntax is based on your impl
if err := goovn.Get(logicalRouter, "lr1"); err != nil {
  return err
}

lrsr := &goovn.LogicalRouterStaticRoute{
       // list of fields here etc....
}

if err := goovn.Create(lrsr); err != nil {
   return err
 }

// Now i need to mutate the LR fields `static_routes` column to include my new ID
// taking a guess here at an API, but maybe?
if err := goovn.Mutate(logicalRouter.StaticRoutes, lr.UUID); err != nil {
   return err
}
```

We're also keen to batch commands in ovn-k8s for performance reasons. I'm not really sure how we'd handle that with this API. Thoughts?

hzhou8:
I thought it is somehow agreed between several people of us (@amorenoz @vtolstov, and now @dave-tucker ) but I think it is better to be clear to put my thoughts here: I think about this PR as example only, because we will have code generator to generate this kind of code instead of manually maintaining it.

Moreover, I am not sure if we should even maintain the generated code in this repo. I assume user of libovsdb (and OVN), such as ovn-k8s, can generate code in the client code repo itself (probably multiple schema versions, in different folders), and make the version selection configurable there. So I am not really sure yet what exactly should remain in go-ovn. I think it will be more clear to discuss when we have the code generator working properly.

hzhou8:
> @amorenoz is this up-to-date? I can't see the `Create()` and `Get()` implementations you linked in #145.
> 
> I'm also curious on how you plan on handling mutations. A very common use case would be as follows (in pseudo-go)
> 
> ```go
> client := goovn.Client()
> // i need to create a logical router static route and reference it from a logical router
> // i know the logical router name though
> 
> var logicalRouter goovn.LogicalRouter
> // I'm not sure what the Get syntax is based on your impl
> if err := client.Get(logicalRouter, "lr1"); err != nil {
>   return err
> }
> 
> lrsr := &goovn.LogicalRouterStaticRoute{
>        // list of fields here etc....
> }
> 
> if err := client.Create(lrsr); err != nil {
>    return err
>  }
> 
> // Now i need to mutate the LR fields `static_routes` column to include my new ID
> // taking a guess here at an API, but maybe?
> if err := client.Mutate(logicalRouter.StaticRoutes, lr.UUID); err != nil {
>    return err
> }
> ```
> 
> We're also keen to batch commands in ovn-k8s for performance reasons. I'm not really sure how we'd handle that with this API. Thoughts?

Great point. I didn't think this through yet but it seems that we need "named-uuid" support across API calls. Firstly call the "create" API with a "named-uuid" specified, and then use the same "named-uuid" to call the "update" API to mutate the object that uses the newly created object. Need more details though.

dave-tucker:
@amorenoz another use case that this API would need to handle is "delete by mutation".
See; https://github.com/eBay/go-ovn/blob/master/logical_router_static_route.go#L156 as an example.
Here to delete the LRSR, we mutate the `static_routes` field to the LR remove the UUID from it's list and due the entry from the static routes table is automagically cleaned up. IIRC a delete will fail on an object that's referenced in another table.

This might require some parent/child modelling in order for this to work as expected.

dave-tucker:
> So I am not really sure yet what exactly should remain in go-ovn. I think it will be more clear to discuss when we have the code generator working properly.

Agreed we'll know more once we have a working generator,
There is still an amount of boilerplate that would be needed...
Setting up OVSDB connection, Setting up the Monitor and the cache population...
The `Get` method (which is very useful indeed) relies on accessing the cache. 
If all of these could be moved in to `libovsdb` then I'd be happy.

hzhou8:
> @amorenoz another use case that this API would need to handle is "delete by mutation".
> See; https://github.com/eBay/go-ovn/blob/master/logical_router_static_route.go#L156 as an example.
> Here to delete the LRSR, we mutate the `static_routes` field to the LR remove the UUID from it's list and the entry from the static routes table is automagically cleaned up. IIRC a delete will fail on an object that's referenced in another table.
> 
> This might require some parent/child modelling in order for this to work as expected.

Would this work by making two API calls and executing the returned commands together in a single transaction?

amorenoz:
@dave-tucker 
I intended to leave the entire CRUD API for a later PR but, let's just put everything on the table so we can discuss clearly.

With the latest update,

> 
> @amorenoz is this up-to-date? I can't see the Create() and Get() implementations you linked in #145.
> 
> I'm also curious on how you plan on handling mutations. A very common use case would be as follows (in pseudo-go)

Please see https://github.com/eBay/go-ovn/pull/136/files#diff-bde7e540ec82692bc60144ad5ce7df3dc960fe297f090b9d8379391a0def5166

> @amorenoz another use case that this API would need to handle is "delete by mutation".

Also, see https://github.com/eBay/go-ovn/pull/136/files#diff-bde7e540ec82692bc60144ad5ce7df3dc960fe297f090b9d8379391a0def5166 for an example of delete by mutation

amorenoz:
> I thought it is somehow agreed between several people of us (@amorenoz @vtolstov, and now @dave-tucker ) but I think it is better to be clear to put my thoughts here: I think about this PR as example only, because we will have code generator to generate this kind of code instead of manually maintaining it.
> 
> Moreover, I am not sure if we should even maintain the generated code in this repo. I assume user of libovsdb (and OVN), such as ovn-k8s, can generate code in the client code repo itself (probably multiple schema versions, in different folders), and make the version selection configurable there. So I am not really sure yet what exactly should remain in go-ovn. I think it will be more clear to discuss when we have the code generator working properly.

@hzhou8 , maybe I misunderstood our previous discussion, but my approach here is not to automatically generate the entire client code but just the Database Model and keep the library generic enough. With the current state of this PR, the client can use the code generator in cmd/modelgen (still very raw) to generate files such as:


```
// Code generated by "ovsmodel"; DO NOT EDIT.

package modelgen

import (
	goovn "github.com/ebay/go-ovn"
)

var (
	ACLTable goovn.TableName = "ACL"
)

//ACL struct defines an object in ACL table
type ACL struct {
	UUID        string            `ovs:"_uuid"`
	Direction   string            `ovs:"direction"`
	Log         bool              `ovs:"log"`
	Meter       []string          `ovs:"meter"`
	Priority    int               `ovs:"priority"`
	Severity    []string          `ovs:"severity"`
	Action      string            `ovs:"action"`
	ExternalIds map[string]string `ovs:"external_ids"`
	Match       string            `ovs:"match"`
	Name        []string          `ovs:"name"`
}

// Table returns the table name. It's part of the Model interface
func (*ACL) Table() goovn.TableName {
	return ACLTable
}
```

As you can see, it's a very simple type definition that just makes live slightly easier for clients but does not implement any logic. Clients can still:
- Create their own types and use them (maybe defining less fields if not interested in all of them)
- Use their old types (auto-generated or not) with newer versions of go-ovn: simple bug fixing for go-ovn and simple upgrading for clients
- Use the old API, i.e: no need to do any kind of code-generation for ovnimp to be able to compile and use go-ovn

What do you think?
If the preferred way is to auto-generate the client code, I'd take a look at @dave-tucker 's approach at #145

amorenoz:
@dave-tucker 

> If all of these could be moved in to libovsdb then I'd be happy.

+1 for moving the cache down to libovsdb

dave-tucker:
> Please see https://github.com/eBay/go-ovn/pull/136/files#diff-bde7e540ec82692bc60144ad5ce7df3dc960fe297f090b9d8379391a0def5166

@amorenoz Thanks for the update... and you're using `OVNCommands` too which allows for us to batch things in to one transaction. Nice!

```
UpdateFields(Model, []string, ...string) (*OvnCommand, error)
```

An issue here is that you are using the `update` operation vs. the `mutate` operation.
This will replace the contents of the field with whatever we provide.
This is an important difference, because the copy of the object we're updating may be stale when the command is executed... which would cause us to overwrite the data in that field.

I wonder whether it might be better to have:
```
Mutate(from Model, to Model) (*OvnCommand, error)
```
This would compare two models and generate a mutate operation for each field that is different

amorenoz:
> > Please see https://github.com/eBay/go-ovn/pull/136/files#diff-bde7e540ec82692bc60144ad5ce7df3dc960fe297f090b9d8379391a0def5166
> 
> @amorenoz Thanks for the update... and you're using `OVNCommands` too which allows for us to batch things in to one transaction. Nice!
> 
> ```
> UpdateFields(Model, []string, ...string) (*OvnCommand, error)
> ```
> 
> An issue here is that you are using the `update` operation vs. the `mutate` operation.
> This will replace the contents of the field with whatever we provide.
> This is an important difference, because the copy of the object we're updating may be stale when the command is executed... which would cause us to overwrite the data in that field.
> 
> I wonder whether it might be better to have:
> 
> ```
> Mutate(from Model, to Model) (*OvnCommand, error)
> ```
> 
> This would compare two models and generate a mutate operation for each field that is different

Yep, that's a good idea. Definitely doable and very useful. The user will then be able to choose if a full update is required or only a mutation is needed.

hzhou8:
> If the preferred way is to auto-generate the client code, I'd take a look at @dave-tucker 's approach at #145

@amorenoz Yes I still think it is better to generate all the code that can be deduced from schema. Otherwise, it is hard to manage the schema changes in go-ovn while keeping backward comparability. For cache implementation, since it is generic to ovsdb clients, I'd also move it to libovsdb, and with your native binding support it is now easier, right?

amorenoz:
The plan is to add this API to libovsdb directly: https://github.com/ovn-org/libovsdb/pull/78
Closing this one

Issue #138, opened 2021-03-16
ACL set APIs needed for match criteria and logging


anfredette:
ovn-kubernetes requires the ability to modify the match criteria and/or logging configuration for an existing ACL.

Issue #137, opened 2021-03-02
ACL: missing 'name' field support


vgozer:
nameless ACLs matching evens in logs look like this: 
`2021-03-02T19:29:05.415Z|06852|acl_log(ovn_pinctrl0)|INFO|name="<unnamed>", verdict=drop, severity=debug: ...`
which is not good ...
trivial to add 'name' field support, but this will change the API signature for ACLAddEntity (extra parameter). Can keep the signature and pass the name in external_ids, but it's ugly... adding a new one (ACLAddEntityNamed) seems like an overkill.
Opinions?

anfredette:
I don't have a problem extending the ACLAddEntity API.  This API is fairly new, so it's probably not widely used.  I currently have a draft PR open on ovn-kubernetes that uses it, but I need to add some additional APIs to go-ovn before it can be merged anyway.

ovn-kubernets uses ACL name a lot, so I've been thinking about how to handle it.  We could change ovn-kubernetes to not use ACL name, stash the ACL name in an external-id, or add the support.  If we do the external-id option, we'd run the risk of using duplicate names.

It's not entirely trivial, but wouldn't be very hard.  We'd need to extend the check for existing ACLs to also consider name, and it would be an "OR" check.  So, if an ACL exists with the same (direction, match and priority) OR (same name), we'd need to reject the ADD.  I think we'd also want to avoid looping through the existing ACLs twice, so I think we might want to replace the getACLUUIDByRow() call with a function that does the "OR" check.

hzhou8:
Adding columns in OVSDB schema is expected over time, but creates problems for go-ovn lib. If we want to keep the backward compatibility, we may have to add new APIs, probably with suffix V2/V3/..., which is not ideal.

I hope this can be solved eventually by API auto-generation. One approach is to generate APIs for different versions of schema to different packages, and let client decide which version to use. I am not sure if there is a better way, say, only generate the APIs for the required version of schema. The C IDL is generated this way at compile time of the client code, but for GO dependency mechanism, not sure if it is possible. cc @amorenoz @vtolstov (we have been discussing this for the native bindings API effort)

For this case for now, I think updating ACLAddEntity signature directly may be ok as mentioned by @anfredette , since it is quite new, if no other objections.

vgozer:
@hzhou8 @anfredette  ACLAddEntity(..., name, ...) would work for me, in the interim. can it be coupled with ACLDelEntity(..., name, ...)?

anfredette:
[...]

PR #139, opened 2021-03-16
Extend ACL API to support changing match critera and logging for an existing ACL

anfredette:
- Support setting ACL name in ACLAddEntity()
- Add ACL set support for ACL name, match and logging for an existing ACL
- Use ACL UUID to identify ACL in ACLDelEntity()

Addresses Issues #137 and #138

### DIFF
--- a/client.go
+++ b/client.go
@@ -73,15 +73,21 @@ type Client interface {
 	LSLBList(ls string) ([]*LoadBalancer, error)
 
 	// Add ACL to entity (PORT_GROUP or LOGICAL_SWITCH)
-	ACLAddEntity(entityType EntityType, entity, direct, match, action string, priority int, external_ids map[string]string, logflag bool, meter string, severity string) (*OvnCommand, error)
+	ACLAddEntity(entityType EntityType, entityName, aclName, direct, match, action string, priority int, external_ids map[string]string, logflag bool, meter, severity string) (*OvnCommand, error)
 	// Deprecated in favor of ACLAddEntity(). Add ACL to logical switch.
 	ACLAdd(ls, direct, match, action string, priority int, external_ids map[string]string, logflag bool, meter string, severity string) (*OvnCommand, error)
+	// Set name for ACL
+	ACLSetName(aclUUID, aclName string) (*OvnCommand, error)
+	// Set match criteria for ACL
+	ACLSetMatch(aclUUID, newMatch string) (*OvnCommand, error)
+	// Set logging for ACL
+	ACLSetLogging(aclUUID string, newLogflag bool, newMeter, newSeverity string) (*OvnCommand, error)
 	// Delete acl from entity (PORT_GROUP or LOGICAL_SWITCH)
-	ACLDelEntity(entityType EntityType, entity, direct, match string, priority int, external_ids map[string]string) (*OvnCommand, error)
+	ACLDelEntity(entityType EntityType, entityName, aclUUID string) (*OvnCommand, error)
 	// Deprecated in favor of ACLDelEntity(). Delete acl from logical switch
 	ACLDel(ls, direct, match string, priority int, external_ids map[string]string) (*OvnCommand, error)
 	// Get all acl by entity
-	ACLListEntity(entityType EntityType, entity string) ([]*ACL, error)
+	ACLListEntity(entityType EntityType, entityName string) ([]*ACL, error)
 	// Deprecated in favor of ACLListEntity(). Get all acl by logical switch
 	ACLList(ls string) ([]*ACL, error)
 
@@ -620,16 +626,28 @@ func (c *ovndb) LBList() ([]*LoadBalancer, error) {
 	return c.lbListImp()
 }
 
-func (c *ovndb) ACLAddEntity(entityType EntityType, entity, direct, match, action string, priority int, external_ids map[string]string, logflag bool, meter string, severity string) (*OvnCommand, error) {
-	return c.aclAddImp(entityType, entity, direct, match, action, priority, external_ids, logflag, meter, severity)
+func (c *ovndb) ACLAddEntity(entityType EntityType, entityName, aclName, direct, match, action string, priority int, external_ids map[string]string, logflag bool, meter, severity string) (*OvnCommand, error) {
+	return c.aclAddImp(entityType, entityName, aclName, direct, match, action, priority, external_ids, logflag, meter, severity)
 }
 
 func (c *ovndb) ACLAdd(ls, direct, match, action string, priority int, external_ids map[string]string, logflag bool, meter string, severity string) (*OvnCommand, error) {
-	return c.aclAddImp(LOGICAL_SWITCH, ls, direct, match, action, priority, external_ids, logflag, meter, severity)
+	return c.aclAddImp(LOGICAL_SWITCH, ls, "", direct, match, action, priority, external_ids, logflag, meter, severity)
 }
 
-func (c *ovndb) ACLDelEntity(entityType EntityType, entity, direct, match string, priority int, external_ids map[string]string) (*OvnCommand, error) {
-	return c.aclDelImp(entityType, entity, direct, match, priority, external_ids)
+func (c *ovndb) ACLSetName(aclUUID, aclName string) (*OvnCommand, error) {
+	return c.aclSetNameImp(aclUUID, aclName)
+}
+
+func (c *ovndb) ACLSetMatch(aclUUID, newMatch string) (*OvnCommand, error) {
+	return c.aclSetMatchImp(aclUUID, newMatch)
+}
+
+func (c *ovndb) ACLSetLogging(aclUUID string, newLogflag bool, newMeter, newSeverity string) (*OvnCommand, error) {
+	return c.aCLSetLoggingImp(aclUUID, newLogflag, newMeter, newSeverity)
+}
+
+func (c *ovndb) ACLDelEntity(entityType EntityType, entityName, aclUUID string) (*OvnCommand, error) {
+	return c.aclDelUUIDImp(entityType, entityName, aclUUID)
 }
 
 func (c *ovndb) ACLDel(ls, direct, match string, priority int, external_ids map[string]string) (*OvnCommand, error) {
--- a/acl.go
+++ b/acl.go
@@ -23,6 +23,7 @@ import (
 // ACL ovnnb item
 type ACL struct {
 	UUID       string
+	Name       string
 	Action     string
 	Direction  string
 	Match      string
@@ -145,7 +146,7 @@ func (odbi *ovndb) getACLUUIDByRow(entityType EntityType, entity string, row OVN
 	return "", ErrorNotFound
 }
 
-func (odbi *ovndb) aclAddImp(entityType EntityType, entity, direct, match, action string, priority int, external_ids map[string]string, logflag bool, meter, severity string) (*OvnCommand, error) {
+func (odbi *ovndb) aclAddImp(entityType EntityType, entityName, aclName, direct, match, action string, priority int, external_ids map[string]string, logflag bool, meter, severity string) (*OvnCommand, error) {
 	var table string
 
 	switch entityType {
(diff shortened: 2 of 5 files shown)


hzhou8 (changes requested):
Thanks for the change. Could you explain why these APIs are needed? Can we achieve the same by deleting the old and adding new ACLs? The user can do both operations in a single transaction, which for my understanding would result in the same outcome in terms of both feature and performance. Did I misunderstand it?

anfredette:
Thanks for your comment.

These APIs are implementing functionality currently supported by the ovn-nbctl database commands and used by ovn-kubernetes. ACLSetMatchEntity() would get called to update the gress policy when an address set is added to or deleted from a pod. ACLSetLoggingEntity() would get called when the logging is enabled/disabled or severity is changed.

I agree that the end state would be equivalent if the user deleted the existing ACL and then re-added the ACL with the new information. This would require more work in ovn-k because we don't currently have all the necessary info to recreate the ACL in the place where it's needed. I haven't tried to work it out yet, but I'm assuming it's possible.

Intuitively, it would seem to have less of a performance impact to update a field in a row than it would to delete the row and re-create it. But, I can't comment on the real performance impact.

I also wonder whether deleting an ACL and then re-adding it (even if it's in the same transaction) would impact the data plane more if the ACL was deleted and readded vs. updated. E.g., is it possible for any packets to slip through while it's being deleted and readded?

I also wonder why these set APIs are different from other the other set APIs that we have.

hzhou8:
> Thanks for your comment.
> 
> These APIs are implementing functionality currently supported by the ovn-nbctl database commands and used by ovn-kubernetes. ACLSetMatchEntity() would get called to update the gress policy when an address set is added to or deleted from a pod. ACLSetLoggingEntity() would get called when the logging is enabled/disabled or severity is changed.
> 
> I agree that the end state would be equivalent if the user deleted the existing ACL and then re-added the ACL with the new information. This would require more work in ovn-k because we don't currently have all the necessary info to recreate the ACL in the place where it's needed. I haven't tried to work it out yet, but I'm assuming it's possible.
> 
In fact to call the "set" API you already have to provide most information of an ACL just so that it can be identified. Maybe the only thing you don't provide is for ACLSetMatchEntity() you don't need the log/meter part. If you want to keep the original log settings when updating an ACL's match condition, and assume you don't know what log settings should it be when you try to update the ACL, then one approach is to Get the ACL (which happens locally) first so you know all the existing information and then do a single transaction with "delete + add".

> Intuitively, it would seem to have less of a performance impact to update a field in a row than it would to delete the row and re-create it. But, I can't comment on the real performance impact.

Intuitively yes, but in fact, OVN has to regenerate the flows for the changed ACL, which costs the same as if it is deleted and added back. If in the future this can be optimized in OVN of course we can look into ways to optimize it. (I hope with the new API framework introduced by https://github.com/eBay/go-ovn/pull/136 and related work this would be easily supported in general)

> 
> I also wonder whether deleting an ACL and then re-adding it (even if it's in the same transaction) would impact the data plane more if the ACL was deleted and readded vs. updated. E.g., is it possible for any packets to slip through while it's being deleted and readded?
> 

This shouldn't be worried as long as the operations are performed in a single transaction, which in turn ensures OVS changes happens atomically.

> I also wonder why these set APIs are different from other the other set APIs that we have.

For other existing "set" APIs, their interfaces are more straightforward - they usually provides "name" as an identifier. For ACL, since we are not using "name" (which is a field added in a late version of the schema), the API would look really scary. And more importantly, if we continue this way there can be more and more "set" APIs added - in theory each field of each table requires a "set" API, which is a huge burden for maintenance. For this specific case, since I don't see real benefit of providing the new APIs instead of using "delete + add", for the reason I mentioned regarding the OVN implementation behind, I am hesitating about adding these APIs. But I am still open to opinions. Let me know if you still strongly feel it is needed.

anfredette:
Sorry for the delay, but I wanted to run this by someone else from the ovn-kubernetes team before I responded.  So, I discussed it with @trozet this morning.  He is in agreement with me that it's preferable to use sets in these cases (and probably others).  I believe that the setting of any attribute should be fair game if there's a use case for it.  Though I don't think we need to arbitrarily add sets for everything.  In many cases, they are not needed.  Objects get created with everything they need and are never modified until they are deleted.  In other cases, it makes more sense to have an add/delete API such as with ports being added and deleted from port groups as pods get created and deleted.

Some of the reasoning/concerns (not necessarily in priority order):
1. Consistent with the ovn-nbctl API 
2. Less impact to the code being migrated from the ovn-nbctl API to the go-ovn API.
3. Less work for the user code than deleting and recreating.
4. Less work in the go-ovn client API code
 - with the delete/add sequence, we'd end up searching the list of ACLs twice - once to find the one to delete and another time to confirm that it doesn't already exist on the add.
 - According to cloc, the delete/add sequence is 122 lines of code, while SetMatch and SetLog are 27 and 37 lines of code, respectively.
 - If we had to also do a get first it would be more.
5. We don't know what the impact on the rest of the system will be (e.g., ovn northd, sbdb, controllers).
6. The UUID for the object will change unnecessarily (does anyone care?) 
7. What would happen to dependent non-root rows if their reference in a root table is deleted?  Are they preserved as long as everything happens in a single transaction?

I made a few more comments below.

> In fact to call the "set" API you already have to provide most information of an ACL just so that it can be identified. Maybe the only thing you don't provide is for ACLSetMatchEntity() you don't need the log/meter part. If you want to keep the original log settings when updating an ACL's match condition, and assume you don't know what log settings should it be when you try to update the ACL, then one approach is to Get the ACL (which happens locally) first so you know all the existing information and then do a single transaction with "delete + add".

The missing fields are:
- ACLSetMatchEntity: action, external_ids, logflag, meter, severity
- ACLSetLoggingEntity: action, external_ids

There are solutions, if necessary.

> 
> > Intuitively, it would seem to have less of a performance impact to update a field in a row than it would to delete the row and re-create it. But, I can't comment on the real performance impact.
> 
> Intuitively yes, but in fact, OVN has to regenerate the flows for the changed ACL, which costs the same as if it is deleted and added back. If in the future this can be optimized in OVN of course we can look into ways to optimize it. (I hope with the new API framework introduced by #136 and related work this would be easily supported in general)
> 

I agree that #136 will help, but it may be a while until it's fully usable.

> > I also wonder why these set APIs are different from other the other set APIs that we have.
> 
> For other existing "set" APIs, their interfaces are more straightforward - they usually provides "name" as an identifier. For ACL, since we are not using "name" (which is a field added in a late version of the schema), the API would look really scary. And more importantly, if we continue this way there can be more and more "set" APIs added - in theory each field of each table requires a "set" API, which is a huge burden for maintenance. For this specific case, since I don't see real benefit of providing the new APIs instead of using "delete + add", for the reason I mentioned regarding the OVN implementation behind, I am hesitating about adding these APIs. But I am still open to opinions. Let me know if you still strongly feel it is needed.

Right, because we don't support names or UUIDs for ACLs, it's necessary to provide the entity, direction, match, and priority to uniquely identify the ACL, so it's not as clean as some of the others.  On a related note, @vgozer has requested that "name" be added to the ACL APIs in https://github.com/eBay/go-ovn/issues/137.  I'd be interested to hear your opinion on it.

As mentioned above, I don't think we necessarily need a set API for every field, but if someone has a use case and offers an implementation, I'd support them.

Thanks,
Andre

hzhou8:
@anfredette thanks for the points. I agree with you "set" APIs are useful and harmless.

> Right, because we don't support names or UUIDs for ACLs, it's necessary to provide the entity, direction, match, and priority to uniquely identify the ACL, so it's not as clean as some of the others. On a related note, @vgozer has requested that "name" be added to the ACL APIs in #137. I'd be interested to hear your opinion on it.
> 

Do you think it is reasonable to support "name" in ACL APIs first and then add "set" APIs only using the "name"? Or do we need two sets of "set" APIs: the ones using "name" and the one using "match + priority + direction"?

anfredette:
> Do you think it is reasonable to support "name" in ACL APIs first and then add "set" APIs only using the "name"? Or do we need two sets of "set" APIs: the ones using "name" and the one using "match + priority + direction"?

I think it would be best to just have one set of APIs. I’d like to confirm with everyone how ACL name will work. For example, is it required?  I’m out today. I’ll follow up on Monday with comments on #139. Thanks!

anfredette:
I've pushed a patch updating the client API code per discussion on Issue #137.  I am requesting feedback before fully implementing the changes.  Thanks, Andre

client.go
```diff
@@ -73,19 +73,21 @@ type Client interface {
...
 	// Deprecated in favor of ACLAddEntity(). Add ACL to logical switch.
 	ACLAdd(ls, direct, match, action string, priority int, external_ids map[string]string, logflag bool, meter string, severity string) (*OvnCommand, error)
-	// Set (change) match criteria for ACL
-	ACLSetMatchEntity(entityType EntityType, entity, direct, oldMatch, newMatch string, priority int) (*OvnCommand, error)
-	// Set (change) logging for ACL
-	ACLSetLoggingEntity(entityType EntityType, entity, direct, match string, priority int, newLogflag bool, newMeter, newSeverity string) (*OvnCommand, error)
+	// Set name for ACL
+	ACLSetName(aclUUID string, aclName string) (*OvnCommand, error)
+	// Set match criteria for ACL
+	ACLSetMatchEntity(entityType EntityType, entityUUID, aclUUID, newMatch string) (*OvnCommand, error)
```
hzhou8, line 82: I think entity type and name/id are not needed for the "set" APIs if ACL uuid is used.

client.go
```diff
@@ -73,19 +73,21 @@ type Client interface {
 	LSLBList(ls string) ([]*LoadBalancer, error)
 
 	// Add ACL to entity (PORT_GROUP or LOGICAL_SWITCH)
-	ACLAddEntity(entityType EntityType, entity, direct, match, action string, priority int, external_ids map[string]string, logflag bool, meter string, severity string) (*OvnCommand, error)
+	ACLAddEntity(entityType EntityType, entityUUID, aclName, direct, match, action string, priority int, external_ids map[string]string, logflag bool, meter, severity string) (*OvnCommand, error)
```
hzhou8, line 76: Why entityUUID is preferred than entityName? I know this has been a big discussion on slack. I replied there as well.

client.go
anfredette: We ensure that the 3-tuple <direction, match, priority> is unique for a given entity.  If we allow the user to change one of these items, we need to make sure that the new 3-tuple does not conflict with another ACL on the same entity.  The ACL record does not contain the entity on which it was created.  So, if we don't pass in the entity, the code would need to potentially loop through every ACL on every entity to find the entity on which the ACL was created before it could confirm that the new 3-tuple was still unique for that entity.

client.go
anfredette: I'll respond in slack so we can keep the discussion in one place and then comment back here when there is a decision.

client.go
hzhou8: I would leave this check to the client instead of in go-ovn API. When using aclUUID it is natural to assume this task on client side. In fact even if entity is passed in, go-ovn API can't guarantee the uniqueness because the same ACL may be used by multiple entities, and this API can only pass in one of them.

So I'd make the "Set" API simple (and efficient). If needed, a Get (or ListBy) API (as discussed earlier) can be provided for querying if certain ACLs existed, before calling the "Set" API.

anfredette: > I would leave this check to the client instead of in go-ovn API. When using aclUUID it is natural to assume this task on client side. In fact even if entity is passed in, go-ovn API can't guarantee the uniqueness because the same ACL may be used by multiple entities, and this API can only pass in one of them.

I don't think the go-ovn API currently allows a given ACL to be associated with more than one entity.  I think it can be done via the ovn-nbctl database API, but I don't think we do that in ovn-kubernetes. So, it seems possible that at some point, it could happen that an ACL is associated with more than one entity.

> 
> So I'd make the "Set" API simple (and efficient). If needed, a Get (or ListBy) API (as discussed earlier) can be provided for querying if certain ACLs existed, before calling the "Set" API.

That's fine.

anfredette:
This PR is ready for review.

anfredette:
I just added one more negative test case.  Now it's really ready for review :-)

hzhou8 (approved):

hzhou8:
Thanks @anfredette

PR #142, opened 2021-03-30
set/delete key-values for OvsMap columns

vgozer:
@hzhou8 @anfredette FYI, this is a replacement for #135. Reworked per suggestions (see review comments). LSP type setting not in scope, will submit a separate PR for that.

### DIFF
--- a/client.go
+++ b/client.go
@@ -248,6 +248,14 @@ type Client interface {
 
 	// GetSchema() returns ovn-db schema
 	GetSchema() libovsdb.DatabaseSchema
+
+	// AuxKeyValSet() sets keys/values for a column of OvsMap type, e.g., 'external_ids', 'other_config'.
+	AuxKeyValSet(table string, row string, auxCol string, kv map[string]string) (*OvnCommand, error)
+	// AuxKeyValDel() removes keys/values for a column of OvsMap type, e.g., 'external_ids', 'other_config'.
+	// special value of 'nil' removes the given key regardless of its value
+	AuxKeyValDel(table string, row string, auxCol string, kv map[string]*string) (*OvnCommand, error)
+	// AuxKeyDel() removes specified keys from a column of OvsMap type
+	AuxKeyDel(table string, row string, auxCol string, keys ...string) (*OvnCommand, error)
 }
 
 var _ Client = &ovndb{}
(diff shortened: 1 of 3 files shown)


ovnimp.go
```diff
@@ -384,3 +384,140 @@ func (odbi *ovndb) optionalStringFieldToPointer(fieldValue interface{}) *string
...
+	}
+
+	operations := []libovsdb.Operation{operation}
+	return &OvnCommand{operations, odbi, make([][]map[string]interface{}, len(operations))}, nil
+}
+
+func (odbi *ovndb) auxKeyValDel(table string, row string, auxCol string, kv map[string]*string) (*OvnCommand, error) {
+	ovnRow := make(OVNRow)
+	ovnRow["name"] = row
+	if len(odbi.getRowUUID(TableLogicalSwitch, ovnRow)) == 0 {
```
hzhou8, line 445: Similar to above, since we do a lookup here already, we can save the uuid and use it for the "condition" in the OVSDB transaction.

hzhou8 (changes requested):

client.go
```diff
@@ -248,6 +248,14 @@ type Client interface {
 
 	// GetSchema() returns ovn-db schema
 	GetSchema() libovsdb.DatabaseSchema
+
+	// AuxKeyValSet() sets keys/values for a column of OvsMap type, e.g., 'external_ids', 'other_config'.
+	AuxKeyValSet(table string, row string, auxCol string, kv map[string]string) (*OvnCommand, error)
```
hzhou8, line 253: For the argument "row", in this PR implementation it is expecting "name" (instead of UUID), so maybe it is better to make it clear in the argument naming, e.g. "rowName".

(It is also possible to use UUID, and the "name" v.s. "UUID" discussion is still on going for other APIs.)

hzhou8 (changes requested):

client.go
vgozer: ack. will fix.

client.go
```diff
@@ -248,6 +248,14 @@ type Client interface {
...
+	// AuxKeyValDel() removes keys/values for a column of OvsMap type, e.g., 'external_ids', 'other_config'.
+	// special value of 'nil' removes the given key regardless of its value
+	AuxKeyValDel(table string, row string, auxCol string, kv map[string]*string) (*OvnCommand, error)
+	// AuxKeyDel() removes specified keys from a column of OvsMap type
```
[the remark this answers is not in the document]
vgozer: @hzhou8 it's only for convenience, since normal maps delete semantics is delete(key). i can remove it, no problem.

ovnimp.go
```diff
@@ -384,3 +384,140 @@ func (odbi *ovndb) optionalStringFieldToPointer(fieldValue interface{}) *string
...
+		}
+	}
+
+	auxMap, err := libovsdb.NewOvsMap(mergedMap)
+	if err != nil {
+		return nil, err
+	}
+	ovnRow[auxCol] = auxMap
+
+	condition := libovsdb.NewCondition("name", "==", row)
```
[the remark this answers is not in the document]
vgozer: ack. will change.

ovnimp.go
vgozer: ditto

PR #144, opened 2021-04-01
Add Options To Logical Router Static Route

dave-tucker:
Signed-off-by: Dave Tucker <dave@dtucker.co.uk>

### DIFF
--- a/client.go
+++ b/client.go
@@ -113,7 +113,7 @@ type Client interface {
 	LRPList(lr string) ([]*LogicalRouterPort, error)
 
 	// Add LRSR with given ip_prefix on given lr
-	LRSRAdd(lr string, ip_prefix string, nexthop string, output_port *string, policy *string, external_ids map[string]string) (*OvnCommand, error)
+	LRSRAdd(lr string, ip_prefix string, nexthop string, output_port *string, policy *string, options map[string]string, external_ids map[string]string) (*OvnCommand, error)
 	// Delete LRSR with given ip_prefix, nexthop, outputPort and policy on given lr
 	LRSRDel(lr string, prefix string, nexthop, outputPort, policy *string) (*OvnCommand, error)
 	// Delete LRSR by uuid given lr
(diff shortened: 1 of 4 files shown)


logical_router_static_route.go
```diff
@@ -172,6 +180,7 @@ func (odbi *ovndb) rowToLogicalRouterStaticRoute(uuid string) *LogicalRouterStat
 		UUID:       uuid,
 		IPPrefix:   cacheLogicalRouterStaticRoute.Fields["ip_prefix"].(string),
 		Nexthop:    cacheLogicalRouterStaticRoute.Fields["nexthop"].(string),
+		Options:    cacheLogicalRouterStaticRoute.Fields["options"].(libovsdb.OvsMap).GoMap,
```
hzhou8, line 183: With older OVN NB schema this code would panic. We need to add a check here to make sure the field exists (not nil).

client.go
```diff
@@ -113,7 +113,7 @@ type Client interface {
 	LRPList(lr string) ([]*LogicalRouterPort, error)
 
 	// Add LRSR with given ip_prefix on given lr
-	LRSRAdd(lr string, ip_prefix string, nexthop string, output_port *string, policy *string, external_ids map[string]string) (*OvnCommand, error)
+	LRSRAdd(lr string, ip_prefix string, nexthop string, output_port *string, policy *string, options map[string]string, external_ids map[string]string) (*OvnCommand, error)
```
hzhou8, line 116: This would break existed client code, but I don't have a better suggestion for dealing with the DB schema change. With the effort of automatic native bindings enhancement in libovsdb by @amorenoz , I hope this can be handled in a better way with code generation. @vtolstov @amorenoz @vgozer do you have any concerns here for updating API signature in go-ovn for now?

hzhou8 (changes requested):

client.go
vgozer: @hzhou8 breaking API (usually a bad thing) is not a concern for me, I can't use the existing method because of its limitations. If this is a concern for others and we don't want to introduce a new method with the same functionality (e.g., LRSRaddNew() - ughh), we could hack the current method to keep the existing signature. e.g., stuff a single map[string]string with both external_ids and options K/Vs, using 'option_' prefix + key name for the latter, so we could build the corresponding column's OvsMap (external_ids key names stored as is, options map keys stripped from 'option_' prefix).  This is also ugly, probably even worse than a new method name. But it won't break the API. Someone would have to remember to clean up later :) In general, we need API versioning/auto-binding, like you said - until then, things like these will continue to haunt us.

hzhou8: Thanks @vgozer for confirming. So I will merge it as is when the other comment is addressed.

PR #145, opened 2021-04-05
(RFC) WIP: Use Go Generate instead of ORM

dave-tucker:
This is a potential alternative to #136 (or it could be used in conjunction with it) and builds upon the work that @amorenoz did for adding bindings to libovsdb.

An ORM API is mainly useful when you are storing abstract data in database. You would create your structs in Go, and the ORM API would deal with creating tables, mapping go types to database types and vice versa etc...

However, as are dealing with data that is described in a Schema having ORM is, in my opinion, less useful.
Our structs in Go, need to mirror that schema exactly.

Therefore, I propose that instead of using an ORM API, we simply us Golang's code generation to generate the bindings.
This has got a couple of advantages:

1. It relies on a very minor change to `libovsdb` to make a method public
2. It would be easy to extend the tool to generate multiple packages (e.g `v4`, v5` etc..) for handling differing schema versions should we want to minimise the impact of API changes for library consumers
3. It removes human error, by generating all tables/fields of the schema automatically

Here is an example of what is generated:

```
/* THIS CODE WAS AUTOGENERATED BY GENFRONSCHEMA */
/* DO NOT EDIT */

package ovn_northbound

import (
	"fmt"
	"reflect"
)

const (
	ACLTableName = "ACL"
)

type ACL struct {
	UUID        string
	Match       string
	Severity    []string
	Meter       []string
	Name        []string
	Direction   string
	Action      string
	Log         bool
	ExternalIds map[string]string
	Priority    int
}

func (odbi *ovndb) rowToACL(uuid string) (*ACL, error) {
	s := &ACL{
		UUID: uuid,
	}
	cache, ok := odbi.cache[ACLTableName][uuid]
	if !ok {
		return s, fmt.Errorf("object does not exist in cache")
	}
	sPtr := reflect.ValueOf(s)
	if sPtr.Type().Kind() != reflect.Ptr {
		return s, fmt.Errorf("not a struct ptr")
	}
	obj := reflect.Indirect(sPtr)

	if v, ok := cache.Fields["match"]; ok {
		err := odbi.colToField("match", v, obj.FieldByName("Match"))
		if err != nil {
			return s, err
		}
	}

	if v, ok := cache.Fields["severity"]; ok {
		err := odbi.colToField("severity", v, obj.FieldByName("Severity"))
		if err != nil {
			return s, err
		}
	}

	if v, ok := cache.Fields["meter"]; ok {
		err := odbi.colToField("meter", v, obj.FieldByName("Meter"))
		if err != nil {
			return s, err
		}
	}

	if v, ok := cache.Fields["name"]; ok {
		err := odbi.colToField("name", v, obj.FieldByName("Name"))
		if err != nil {
			return s, err
		}
	}

	if v, ok := cache.Fields["direction"]; ok {
		err := odbi.colToField("direction", v, obj.FieldByName("Direction"))
		if err != nil {
			return s, err
		}
	}

	if v, ok := cache.Fields["action"]; ok {
		err := odbi.colToField("action", v, obj.FieldByName("Action"))
		if err != nil {
			return s, err
		}
	}

	if v, ok := cache.Fields["log"]; ok {
		err := odbi.colToField("log", v, obj.FieldByName("Log"))
		if err != nil {
			return s, err
		}
	}

	if v, ok := cache.Fields["external_ids"]; ok {
		err := odbi.colToField("external_ids", v, obj.FieldByName("ExternalIds"))
		if err != nil {
			return s, err
		}
	}

	if v, ok := cache.Fields["priority"]; ok {
		err := odbi.colToField("priority", v, obj.FieldByName("Priority"))
		if err != nil {
			return s, err
		}
	}

	return s, nil
}

// TODO: Implement Add/Delete functions and the inverse of rowToX

```

NOTE: I've currently set this to generate in to a `ovn_northbound` package.
Once i've implemented the remaining functions, I plan to generate these files in the repository root, adjust `client.go` appropriately and hopefully all the tests would still pass.

Should #136 be preferred by the maintainers, it would be relatively easy to generate all of the structs automatically using a small program like this.


client.go
```diff
@@ -353,6 +354,26 @@ func (c *ovndb) reconnect() {
 	}()
 }
 
+// colToField uses reflection to convert a value from ovsdb to a provided struct field
+func (odbi *ovndb) colToField(column string, from interface{}, to reflect.Value) error {
+	schema := odbi.GetSchema()
+	tableSchema, ok := schema.Tables[TableACL]
```
amorenoz, line 360: Shouldn't "Table" be a parameter to this function?

amorenoz:
Hi @dave-tucker, thanks for this nice idea.
I thought of doing this as well. Although I think the ORM approach has some useful properties, I agree that some of them might not be urgently needed.

My main concerns are two: 
1) **Try to avoid API explosion,** : try to avoid having LSList, LRGet, .... and instead have a generic List(). Same goes for the callback API (OnLogicalRouterCreated()). The current limitation makes the library unusable for many monitoring purposes.


I achieve this by creating a [Model](https://github.com/eBay/go-ovn/blob/c9d068664ff50ec46a803bb73ba77559ed0693f3/model.go#L28) interface that all types implement and that can be used to properly cast the struct on the callback function [as in the example](https://github.com/eBay/go-ovn/blob/c9d068664ff50ec46a803bb73ba77559ed0693f3/examples/orm/main.go#L26-L32).

Similarly, generic [List()](https://github.com/eBay/go-ovn/blob/c9d068664ff50ec46a803bb73ba77559ed0693f3/ovnimp_orm.go#L71), [Get()](https://github.com/amorenoz/go-ovn/blob/6707d7ad0208e64a87e77a40078ee2062d866788/ovnimp_orm.go#L52), [Create()](https://github.com/amorenoz/go-ovn/blob/6707d7ad0208e64a87e77a40078ee2062d866788/ovnimp_orm.go#L52)... functions can be added using only one parameter (typically, the pointer to the object to be filled).

This can, of course, be done using your approach (e.g: adding the table name to the functions and using `interface{}` as the return values). Just putting it out there for discussion and feedback :)

2) Enable the generated code to live **outside** go-ovn. 

My fear was (is) that if go-ovn ties itself to a particular version of ovn (schema), we might be creating new problems for users that want to use the library for other versions. We could decide to only track stable ovn branches but backporting fixes to the ovn-tracking branches would increase maintenance cost. By having the generated code outside go-ovn, it's the client (e.g: ovn-kubernetes) the one who decides what version of go-ovn and what version of ovn (i.e: what schema) to use. I wrote a small ORM model generator in https://github.com/amorenoz/ovnmodel to test my experiments that demonstrates this (the generator code itself needs a complete cleanup but it demonstrates the point).

What do you think?

amorenoz:
Also interested in @palanik1's take

client.go
dave-tucker: Yes hence the `WIP` 😂 
I tried this code out on the ACL resource and forgot to factor that out, good catch

hzhou8:
Thanks @dave-tucker ! Code generation for ovsdb schema is what we have always wanted to do, so I definitely agree with the idea, and I think with @amorenoz 's work in libovsdb we are very close. In fact I was thinking the next step after https://github.com/eBay/libovsdb/pull/31 is to have a code generation tool in libovsdb to generate API implementations for a given schema based on the ORM API.

So first of all I think this tool should be in libovsdb. In fact I am not quite sure if there is still a need for go-ovn to exist as a separate repo after that - maybe just a wrapper on top of / managing the generated APIs. I'd like to hear your inputs.

Secondly, I think code generation is not conflict with ORM API. Instead, it can be based on the ORM API. What do you think? @dave-tucker @amorenoz

dave-tucker:
> Try to avoid API explosion, : try to avoid having LSList, LRGet, .... and instead have a generic List(). Same goes for the callback API (OnLogicalRouterCreated()). The current limitation makes the library unusable for many monitoring purposes.

Ack. This won't avoid exploding the API. In fact it will have the opposite effect as we'll automatically have bindings for all the tables 😢 

Another thing that needs to be solved is 

> My fear was (is) that if go-ovn ties itself to a particular version of ovn (schema), we might be creating new problems for users that want to use the library for other versions....

In my experience, you're damned if you do, damned if you don't 😉 

If we push the responsibility to users, we'll need the ability to constrain the schema version in the config... otherwise we run the risk of a user hitting an incompatible OVN schema version at runtime and the library won't work as expected. There may be varying `ovn` versions in the wild depending on which release train of which linux distro you're running on.

On the other hand, we don't have to have loads of branches here... I think (although I haven't totally thought this through) that a library user could choose the major version (i.e `import `github.com/eBay/go-ovn/ovn_northbound/v5/`) and we could (optionally) get fancy and support fallback if the OVN version doesn't meet the required schema level.
_Disclaimer: This is theoretical_

dave-tucker:
> So first of all I think this tool should be in libovsdb. In fact I am not quite sure if there is still a need for go-ovn to exist as a separate repo after that - maybe just a wrapper on top of / managing the generated APIs. I'd like to hear your inputs.

I think the generation tool could live in `libovsdb`. Then `go-ovn` would become little more than the generated structs and some glue code... unless (see comment above) there is a desire to get fancy and handle things like mismatched client versions.

> Secondly, I think code generation is not conflict with ORM API. Instead, it can be based on the ORM API.

Agreed. You can use it with, or without. I really don't have a strong opinion about it. 
I'll go ahead and leave some comments on #145 though

dave-tucker:
Closing this PR for two reasons.

1. This functionality should live in libovsdb
2. Having spent a little more time today playing with this... it's clear the API offered by #136 is much, much better

PR #146, opened 2021-04-07
add LSP type setter

vgozer:
need this is to allow setting of a desired logical switch port type, e.g., localnet

### DIFF
--- a/client.go
+++ b/client.go
@@ -62,6 +62,8 @@ type Client interface {
 	LSPSetAddress(lsp string, addresses ...string) (*OvnCommand, error)
 	// Set port security per lport
 	LSPSetPortSecurity(lsp string, security ...string) (*OvnCommand, error)
+	// Set logical switch port type
+	LSPSetType(lsp string, portType string) (*OvnCommand, error)
 	// Get all lport by lswitch
 	LSPList(ls string) ([]*LogicalSwitchPort, error)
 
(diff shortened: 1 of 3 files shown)


hzhou8:
It is simple and straightforward so I just merged it.

I am not really sure if this will be used at all because it requires a separate transaction after the LSP is created. We need to find a way to support "namedUUID" so that create and update can be done in separate APIs but still can be executed in the same transaction.

vgozer:
@hzhou8 thanks for merging. all points you made bothered me as well, but I could not see a cleaner option - needed a 'localnet' port type, no API, didn't want to change signature of the existing methods. So, adding a simple setter sounded like a compromise - better than exec'ing ovn-nbctl to set type after everything else is done via go-ovn. I wish we had a common signature for methods that will take an attributes map (or entity-specific struct) where we can do any sorts of entity's columns manipulation by adding code to the method's body rather then multiply API methods for each and every attribute.

hzhou8:
@vgozer sorry, I take it back. Actually I forgot that since the API is using port "name" instead of "uuid", the user can just call LSPAdd() and then LSPSetType(), and execute the returned commands together in a single transaction, something like:

cmd1, _ = ovndbapi.LSPAdd("ls1", "lsp1")
cmd2, _ = ovndbapi.LSPSetType("lsp1", "localnet")
ovndbapi.Execute(cmd1, cmd2)

PR #148, opened 2021-04-30
Connect - reconnect cache correctness

fedepaol:
Two fixes here:
- we wipe out the cache when reconnecting (this is because we don't get the "deletion" events from MonitorTables
- we setup the notifier before fetching the db dump and initializing the cache. This avoid missed events in between

### DIFF
--- a/client.go
+++ b/client.go
@@ -314,6 +314,9 @@ func connect(c *ovndb) (err error) {
 	notifier := ovnNotifier{c}
 	ovsdb.Register(notifier)
 
+	// When we connect we initialize the cache, so any deletions
+	// happened while reconnecting are handled correctly.
+	c.cache = make(map[string]map[string]libovsdb.Row)
 	initial, err := c.MonitorTables("")
 	if err != nil {
 		return err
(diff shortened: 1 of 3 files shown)


fedepaol:
cc @dave-tucker & @trozet

trozet:
@hzhou8

trozet (approved):
/lgtm

client.go
```diff
@@ -314,6 +314,12 @@ func connect(c *ovndb) (err error) {
 	notifier := ovnNotifier{c}
 	ovsdb.Register(notifier)
 
+	// The cache is re-initialized. This is because when reconnecting, deletions
+	// are not propagated by MonitorTables and we would risk the cache to conain
+	// elements that are not in the db anymore
+	if reconnecting {
```
hzhou8, line 320: Minor comment here: I think this arg is not needed. We can simply initialize the cache for both cases. We can remove the initialization at L346.

hzhou8 (commented):
Thanks for the fix! Looks good to me except a minor comment.

client.go
fedepaol: Yup, that was the other option I had in mind but I didn't like much the fact that New leaves the object half initialized.
I will change that

client.go
```diff
@@ -349,7 +362,7 @@ func (c *ovndb) reconnect() {
 		log.Printf("%s disconnected. Reconnecting ... \n", c.addr)
 		retry := 0
 		for range ticker.C {
-			if err := connect(c); err != nil {
+			if err := connect(c, true); err != nil {
```
hzhou8, line 365: I think you just forgot to remove the new arg here in the revision.

fedepaol: Yep, sorry. Re-pushed

PR #149, opened 2021-06-02
PortGroup: make group membership APIs idempotent

squeed:
Rather than checking to see if a given port is already in a group, just
always transact. The ovsdb protocol already understands set membership
and will happily allow duplicate adds.

Signed-off-by: Casey Callendrello <cdc@redhat.com>

### DIFF
--- a/port_group.go
+++ b/port_group.go
@@ -117,14 +117,8 @@ func (odbi *ovndb) pgUpdateImp(group string, ports []string, external_ids map[st
 	return &OvnCommand{operations, odbi, make([][]map[string]interface{}, len(operations))}, nil
 }
 
-func (odbi *ovndb) pgAddPortImp(group, port string)  (*OvnCommand, error) {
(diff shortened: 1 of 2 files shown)


squeed:
Note: a few gofmt updates in this PR, just view the diff ignoring whitespace.

hzhou8:
LGTM. @anfredette do you want to take a look?

hzhou8 (approved):

hzhou8:
One concern here is, if this idempotent behavior is abused, it could lead to unnecessary load on the OVSDB server. It would be the user's responsibility not abusing it. Alternatively we can also make it idempotent by checking in the library just like now without sending request to the server, just returning success, but the current implementation of the local check is slow, and the cost is there even for regular operations. So, I am ok for either approach.

anfredette (approved):
I actually debated with myself on whether to fail that case.

LGTM

PR #150, opened 2021-06-30
no need to destroy LSP row when also removing from LS' ports column

girishmg:
it is just sufficient to remove the LSP's UUID from the LS's Ports
column, and this in itself will remove the LSP from the table since
LSP can't be present by itself

@hzhou8 @noah8713 PTAL


hzhou8 (approved):

girishmg:
@noah8713 can you please merge this?

PR #151, opened 2021-07-01
In pgUpdateImp method, check for port slice length instead of checkin…

Pardhakeswar:
…g whether ports slice is nil

If we check for whether port is nil, in this case, there can be possibility of port
not being nil and the port slice len is 0 (if port is created using make([]string, 0, 0)).
Then, we are observing below error
 - Transaction Failed due to an error: syntax error details: expected ["set", <array>] in
   {update Port_Group map[name:<a17069877248180744570> ports:0xc00052c528] [] [] [] 0 [[name == a17069877248180744570]]  }

### DIFF
--- a/port_group.go
+++ b/port_group.go
@@ -82,11 +82,11 @@ func (odbi *ovndb) pgUpdateImp(group string, ports []string, external_ids map[st
 		return nil, ErrorNotFound
 	}
 
-	if ports == nil && external_ids == nil {
+	if len(ports) == 0 && external_ids == nil {
 		return nil, ErrorNoChanges
 	}
 
-	if ports != nil {
+	if len(ports) > 0 {
 		portUUIDs := make([]libovsdb.UUID, 0, len(ports))
 		for _, u := range ports {
 			portUUIDs = append(portUUIDs, stringToGoUUID(u))


Pardhakeswar:
@hzhou8 @girishmg @noah8713 PTAL

hzhou8 (approved):

girishmg:
@Pardhakeswar

NIT: your commit subject is too big.. can you just say.. check for port slice length in pgUpdateImp

girishmg:
@noah8713 can you please merge this?

hzhou8:
Thanks @Pardhakeswar and @girishmg . The change is small and obvious, so I just merged.

trozet:
@girishmg @hzhou8 not sure if you guys care anymore, but I think this commit was wrong. This makes it so that if a user intentionally passes an empty array to update the ports from [a,b,c] to [] it will not work if external_ids is nil:

```
I0307 15:14:04.208929       1 policy.go:804] Processing NetworkPolicy e2e-network-policy-8729/deny-client-a-via-except-cidr-egress-rule to have 0 local pods...
I0307 15:14:04.211656       1 policy.go:813] Setting NetworkPolicy e2e-network-policy-8729/deny-client-a-via-except-cidr-egress-rule to have 0 local pods...
E0307 15:14:04.212115       1 policy.go:901] Failed to set ports in PortGroup for network policy e2e-network-policy-8729/deny-client-a-via-except-cidr-egress-rule: no changes requested
I0307 15:14:04.212138       1 policy.go:904] Done setting NetworkPolicy e2e-network-policy-8729/deny-client-a-via-except-cidr-egress-rule local pods

```
The reason @Pardhakeswar hit the error in his description is because NewOVSSet had a bug that we fixed in our downstream:
https://github.com/openshift/ovn-kubernetes/commit/35677418d2bbfddb6229e1d776bba2064dde646b#diff-88e093886eb91e9ca5f9234d74a5f756c0251d685c141c902a7833d95bec5345R27

```
// NewOvsSet creates a new OVSDB style set from a Go slice
func NewOvsSet(goSlice interface{}) (*OvsSet, error) {
	v := reflect.ValueOf(goSlice)
	if v.Kind() != reflect.Slice {
		return nil, errors.New("OvsSet supports only Go Slice types")
	}

-	var ovsSet []interface{}
+	ovsSet := make([]interface{}, 0, v.Len())
	for i := 0; i < v.Len(); i++ {
		ovsSet = append(ovsSet, v.Index(i).Interface())
	}
```

PR #152, opened 2021-07-30
Remove transaction mutex

trozet:
The transaction mutex is not necessary, because the rpc2 lib and ovsdb
protocol ensure that a cache/db update is received before the
transaction is complete, and every transaction is serialized.

Signed-off-by: Tim Rozet <trozet@redhat.com>


trozet:
@hzhou8 @girishmg

hzhou8 (approved):
LGTM @softwarejl do you have any concerns?

PR #153, opened 2021-09-10
Add back tranmutex

trozet:
Need a RWMutex to protect nil client usage during reconnect. Also includes a workaround for https://bugzilla.redhat.com/show_bug.cgi?id=2000375

### DIFF
--- a/ovnimp.go
+++ b/ovnimp.go
@@ -166,7 +166,8 @@ func (odbi *ovndb) transact(db string, ops ...libovsdb.Operation) ([]libovsdb.Op
 			if i < len(ops) {
 				opsInfo = fmt.Sprintf("%v", ops[i])
 			}
-			return nil, fmt.Errorf("Transaction Failed due to an error: %v details: %v in %s",
+			odbi.close()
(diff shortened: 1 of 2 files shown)


trozet:
@girishmg @hzhou8 @tssurya @dcbw

hzhou8:
Hi @trozet, the problem looks clear but I wonder if this is the right way to fix. There can be different reasons for a transaction to fail - it can be temporary. Are we going to blindly reconnect regardless of the reason?

ovnimp.go
```diff
@@ -164,7 +166,8 @@ func (odbi *ovndb) transact(db string, ops ...libovsdb.Operation) ([]libovsdb.Op
 			if i < len(ops) {
 				opsInfo = fmt.Sprintf("%v", ops[i])
 			}
-			return nil, fmt.Errorf("Transaction Failed due to an error: %v details: %v in %s",
+			odbi.close()
```
hzhou8, line 169: it doesn't trigger reconnect if odbi.reconn is not true.

